PR #5, opened 2019-07-23
Data import fixes

nikolaystanishev:


nikolaystanishev:
[SP20: Make data-hg19 compatible with impala backend.](https://trello.com/c/l2xuM71t/948-sp20-make-data-hg19-compatible-with-impala-backend)

PR #12, opened 2019-07-24
Impala parquet loader

nikolaystanishev:


nikolaystanishev:
[SP20: Add tool for loading parquet files in impala db](https://trello.com/c/sPpAUWCz/953-sp20-add-tool-for-loading-parquet-files-in-impala-db)

PR #15, opened 2019-07-24
DAE config default values

nikolaystanishev:


nikolaystanishev:
[SP20: Add Impala section to DAE config](https://trello.com/c/CA1xE4mV/944-sp20-add-impala-section-to-dae-config)

PR #24, opened 2019-08-01
Fix dae2parquet

nikolaystanishev:


nikolaystanishev:
[SP11: 14858 family in svip probably has no proband](https://trello.com/c/S7YpVw9l/574-sp11-14858-family-in-svip-probably-has-no-proband)

DAE/tools/dae2parquet.py
```diff
@@ -302,7 +302,7 @@ def parse_cli_arguments(dae_config, argv=sys.argv[1:]):
             argv.families, argv.variants,
             family_format=argv.family_format,
             output=argv.output,
-            rows=argv.rows, bucket_index=0
+            bucket_index=0
```
lchorbadjiev: Why?

PR #26, opened 2019-08-02
Split values in big queries

nikolaystanishev:


nikolaystanishev:
[SP20: split variant in members where clause](https://trello.com/c/2JLjKlY4/962-sp20-split-variant-in-members-where-clause)

PR #34, opened 2019-09-11
Some fixes to variants querying

iordanivanov:
Backend's query_variants now gets an explicit set of keyword arguments instead of a
kwargs dict with all keyword arguments; This was done to fix errors
with excess keyword arguments getting passed to the backend

The genomicScores filter's name of the genomic score being filtered is
now converted to lowercase due to the way Impala stores column names

minor jenkins_data update


iordanivanov:
before merging I have to merge master into this branch and make sure other keyword arguments aren't being popped since it shouldn't be needed anymore

PR #36, opened 2019-09-11
Configurable permission denied prompt

iordanivanov:


wdae/wdae/datasets_api/urls.py
```diff
@@ -8,6 +8,9 @@
 
 
 urlpatterns = [
+    url(r'^/deniedPrompt$',
```
nikolaystanishev: this should be snake case
```suggestion
    url(r'^/denied_prompt$',
```

wdae/wdae/datasets_api/tests/test_permission_denied_prompt.py
```diff
@@ -0,0 +1,33 @@
+import os
```
nikolaystanishev: you should consider using user_client in this tests

PR #64, opened 2019-10-03
Use variants/family.py's pedigree loading method in pheno/prepare

iordanivanov:


dae/dae/pheno/prepare/ped_prepare.py
```diff
@@ -66,8 +64,12 @@ def _map_role_column(self, ped_df):
 
         roles = pd.Series(ped_df.index)
         for index, row in ped_df.iterrows():
-            role = mapping.get(row['role'])
-            roles[index] = role.value
+            if type(row['role']) is not Role:
+                role = mapping.get(row['role'])
```
lchorbadjiev: When this could happen?

Role mapping should go in pre-processing phenotype data.

iordanivanov: Yes, this shouldn't happen under normal circumstances - perhaps I should've left in an assertion instead

PR #75, opened 2019-10-21
Fix gene sets download request

nikolaystanishev:


nikolaystanishev:
[Fix gene sets download request](https://trello.com/c/awhSdpRR/1203-fix-gene-sets-download-request)

PR #96, opened 2019-11-19
Config parser work dir change

IvoTod:


IvoTod:
[SP27: Set config work directories to point to config file directory](https://trello.com/c/Iibp9m9g/1291-sp27-set-config-work-directories-to-point-to-config-file-directory)

PR #97, opened 2019-11-20
Family allele best state

IvoTod:


IvoTod:
[SP12: [gpf] restore family allele best state](https://trello.com/c/PeRIG55t/638-sp12-gpf-restore-family-allele-best-state)

PR #99, opened 2019-11-29
Parquet loader

IvoTod:


IvoTod:
[Parquet Dataset Loader](https://trello.com/c/wIpp2toW/1315-parquet-dataset-loader)

dae/dae/backends/impala/loader.py
```diff
@@ -0,0 +1,93 @@
...
+        else:
+            filenames = []
+            for part in self.partitions_to_read:
+                filenames += glob.glob(os.path.join(self.root_dir, part))
+
+            return filenames
+
+    def _parquet_file_iterator(self, filename):
+        table = pq.read_table(filename)
+        df = table.to_pandas()
```
lchorbadjiev: this reads all data from parquet file into memory... strongly prefer to read by row groups and iterator over row groups keeping only one row group into memory...

PR #108, opened 2019-12-18
Dataset details route

IvoTod:


wdae/wdae/datasets_api/views.py
```diff
@@ -59,3 +60,24 @@ def __init__(self):
 
     def get(self, request):
         return Response({'data': self.permission_denied_prompt})
+
+
+class DatasetDetailsView(QueryBaseView):
+    permission_classes = (permissions.IsAdminUser,)
```
iordanivanov: Not sure about this - @lchorbadjiev is this alright?

PR #157, opened 2021-04-23
Security updates

iordanivanov:


IvoTod (dismissed):
Tests should check the internal AuthenticationLog model for changes

Login view has become too large and confusing, code handling time calculations and login attempt tracking/checking can be encapsulated inside class methods

IvoTod (approved):

PR #174, opened 2022-02-01
Add histogram computation capabilities to the grr cli

svetlin-mladenov:
Add a new command `histogram` to `grr_manage`. The command takes as arguments a genomic resource repo and a resource in that repo. It generates histograms for that resource. The histograms are described in the resource `genomic_resource.yaml` file. The output histogram is written in a directory `histograms` under the resource path.

The histogram calculation is done in parallel. The parallelism is based on the contigs - for each config a separate job. The number of jobs to be executed in parallel can be controlled through the `--jobs` command line parameter. The default behavior is to start as many jobs as processors on the machine.

### DIFF
--- a/dae/dae/genomic_resources/genomic_scores.py
+++ b/dae/dae/genomic_resources/genomic_scores.py
@@ -105,23 +105,18 @@ class GenomicScore(abc.ABC):
 
     def fetch_region(self, chrom: str, pos_begin: int, pos_end: int,
                      scores: List[str]) \
-            -> dict[str, Generator[Number, None, None]]:
+            -> Generator[dict[str, Number], None, None]:
         if chrom not in self.get_all_chromosomes():
             raise ValueError(
                 f"{chrom} is not among the available chromosomes.")
 
-        res = {}
-        for scr_id in scores:
-            res[scr_id] = self.iter_lines(chrom, pos_begin, pos_end, scr_id)
-        return res
-
-    def iter_lines(self, chrom: str, pos_begin: int, pos_end: int,
-                   scr_id: str) -> Generator[Number, None, None]:
         line_iter = self._fetch_lines(chrom, pos_begin, pos_end)
         for line in line_iter:
             line_pos_begin, line_pos_end = self._line_to_begin_end(line)
 
-            val = line.get_score_value(scr_id)
+            val = {}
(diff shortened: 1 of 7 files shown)


dae/dae/genomic_resources/genomic_scores.py
```diff
@@ -90,20 +91,51 @@ def _get_header(self):
...
 
     def get_all_chromosomes(self):
         return self.table.get_chromosomes()
 
     def get_all_scores(self):
         return list(self.score_columns)
 
+    def fetch_region(self, chrom: str, pos_begin: int, pos_end: int,
+                     scores: List[str]) \
+            -> dict[str, Generator[Number, None, None]]:
```
lchorbadjiev, line 108: I don't think this return type is a good idea. To have multiple independent generators would mean that we are going to iterate over all records multiple times.

I strongly prefer to have Generator[Dict[str, number], None, None].

dae/dae/genomic_resources/score_statistics.py
```diff
@@ -127,6 +129,131 @@ def add_value(self, value):
...
+        limits = np.iinfo(np.int64)
+        scr_min, scr_max = limits.max, limits.min
+        for i_min, i_max in min_maxes:
+            scr_min = min(scr_min, i_min)
+            scr_max = max(scr_max, i_max)
+
+        logger.info(f"Done calculating min/max for {score_id}.\
+ min={scr_min} max={scr_max}")
+        assert scr_min <= scr_max
+        return scr_min, scr_max
```
lchorbadjiev, line 217: consider re-organizing min/max calculation to become similar to histogram calculation _do_hist/_merge_hist

consider keeping the parallelization into one place - inside build() method.

lchorbadjiev (changes requested):

dae/dae/genomic_resources/genomic_scores.py
svetlin-mladenov: @lchorbadjiev fixed

dae/dae/genomic_resources/score_statistics.py
svetlin-mladenov: @lchorbadjiev fixed

lchorbadjiev (approved):

PR #180, opened 2022-02-24
Phenotype groups

IvoTod:


dae/dae/configuration/schemas/phenotype_data.py
```diff
@@ -17,6 +17,7 @@
 phenotype_data_schema = {
     "enabled": {"type": "boolean", "default": True},
     "name": {"type": "string"},
+    "group_datas": {"type": "list", "schema": {"type": "string"}},
```
lchorbadjiev, line 20: "group_datas" does not sound well. Could you please change it? Something like "phenotype_data_list" or "phenotype_data_group" or ...

lchorbadjiev (approved):

PR #193, opened 2022-05-18
New method to import data: Import Tools

svetlin-mladenov:
This PR includes a new tool called `import_tools`. As input it gets an config yaml file describing the data that needs to be imported.

### DIFF
--- a/dae/dae/import_tools/import_tools.py
+++ b/dae/dae/import_tools/import_tools.py
@@ -204,10 +204,13 @@ class ImportProject():
         # TODO handle inline storage configurations
         return self.import_config["destination"].get("storage_id")
 
-    def has_destination(self):
+    def has_destination(self) -> bool:
+        """
+        Returns True if there is a *destination* section in the import config
+        """
         return "destination" in self.import_config
 
-    def get_row_group_size(self, bucket):
+    def get_row_group_size(self, bucket) -> int:
         return self.import_config.get("parquet_row_group_size", {})\
             .get(bucket.type, 20_000)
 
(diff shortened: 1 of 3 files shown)


dae/dae/import_tools/tests/test_custom_region_length.py
```diff
@@ -0,0 +1,253 @@
+from copy import deepcopy
+import pytest
+import os
+from os.path import join
+from dae.configuration.gpf_config_parser import GPFConfigParser
+from dae.import_tools import import_tools
+import pyarrow.parquet as pq
+
+
+def test_import_task_bin_size(gpf_instance_2019, tmpdir):
```
lchorbadjiev, line 10: this test is huge... consider splitting it into multiple tests... or at least add some comments

dae/dae/import_tools/import_tools.py
```diff
@@ -0,0 +1,381 @@
...
+        res = []
+        for chr in chromosomes:
+            if chr in {"autosomes", "autosomesXY"}:
+                for i in range(1, 23):
+                    res.append(f"{i}")
+                    res.append(f"chr{i}")
+                if chr == "autosomesXY":
+                    for i in ["X", "Y"]:
+                        res.append(f"{i}")
+                        res.append(f"chr{i}")
```
lchorbadjiev, line 338: avoid using `chr` since it is a python function... prefer using `chrom`

lchorbadjiev (changes requested):
Some general notes:

- consider adding documentation strings (could be a one-liner) for public methods in new classes... especially for `ImportProject`, `TaskGraph`, etc

- consider transforming `documentation_notest.txt` into `import_tools` package documentation 

- Jenkins build has failing tests... some of these failing tests could be redundant

PR #227, opened 2022-10-05
YAML configurations

IvoTod:
## Background

We initially changed our configuration format to TOML as a part of our configuration reorganization and clean up. While very functional and simple to use, TOML and its Python support have strange cases and quirks when we try and use it for more complex cases. This is most present with lists of dictionaries. Because of this, TOML also required maintaining a more manual "configuration generator" using templates, as dumping a dictionary into a file would often create harder to read sections. For this reason, we decided to change our configuration format to YAML.

## Aim

The major change in this pull request is the deprecation of TOML configurations and changing the `StudyConfigBuilder` to create YAML files instead.
A few changes were made to the study configuration schema:
* `ped_no_role`, `ped_no_header` and `ped_tags` are now booleans instead of strings expected to be "true" or "false".
* `families_count_show_id` in the common reports family report configuration is now removed.

## Implementation

There may be other places in our projects/pipelines where we prepare configurations for an instance in place using sed, these should be updated for YAML instead of TOML.


lchorbadjiev (approved):
It seems OK.

Before merging:
* Check that the "gpf-e2e" build is working with these changes. Transfer the study configurations used there to the YAML.
* Make sure, that "data-hg38-production" instance configuration works on our internal impala cluster. At the moment there are some errors while validating study configs.

PR #229, opened 2022-10-05
Import tools features

svetlin-mladenov:
## Background

This PR is basically polishing of the import-tools functionality.

## Aim

The changes are: 
 * Log the current stage when running import-tools
 * Minor fixes to the documentation.
 * Error with a nice message if `add_chrom_prefix` (or `del_chrom_prefix`) is set for an set of variants that already have prefix added or removed. This makes working with import-tools a less error prone activity.


lchorbadjiev (approved):
Looks OK.

PR #233, opened 2022-10-10
Summary variants in schema2

### LINKED ISSUE #226: [ENH] support for `family_variants_count`, `seen_as_denovo` and `seen_in_status` attributes of summary variants in impala schema2

At the moment impala schema2 genotype storage does not support storing some of the summary variants attributes:
* `family_variants_count`
* `seen_in_denovo` 
* `seen_in_status`

Import variants in schema2 should support adding these attributes with appropriate values.

**Additional context**
Because these attributes are missing the gene browser does not show the correct family variants counts and colors for displayed summary variants when working with impala schema2 genotype storage.

### LINKED ISSUE #228: [ENH] pytest markers to specify storage type subset to use in an integration test
Currently, the parametrization of integration tests with a genotype storage fixture uses all storage types. We need a way to restrict storage types for a specific test and make it run for a subset of storage types.

Add `pytest` markers for each of the registered genotype storage types:
* `filesytem`
* `impala`
* `impala2`
and adjust the `genotype_storage` fixture to filter injected genotype storages accordingly.

**Describe alternatives you've considered**
No idea of an alternative solution at the moment.

**Additional context**
We want to evolve and enhance our new genotype storages without backporting features to obsolete and deprecated storages.

### LINKED ISSUE #230: [ENH] simplify definition of reference genome and gene models in integration testing
Most integration tests require the definition of a reference genome and gene models. We should have an easy way to define these objects in a testing environment.

Define functions:

* `setup_reference_genome`
* `setup_gene_models`

that simplifies the definition of these objects in a testing environment.

**Additional context**
It would be good if there is a way to pass these definitions when creating a testing GPF instance and import project.

lchorbadjiev:
## Background

The gene browser expects that summary variants have the following additional attributes:

* `seen_in_status`
* `seen_as_denvo`
* `family_variants_count`

These attributes are calculated on the fly in Impala Schema1 genotype storage. 

## Aim

The data import in Schema2 calculates these additional attributes and stores them in summary alleles parquet files.
Integration tests for all these attributes are added, including tests using genotype data groups. Some of the helper functions for setting up integration tests are simplified. GPF instance class constructor is changed to accept prebuild reference genome and gene models.

## Implementation

At the moment, all these attributes are made allele-specific. In Schema 1, they are variant-specific. I am not sure which approach is better. We should revisit this behavior.

Closes #226
Closes #228
Closes #230.


dae/dae/configuration/gpf_config_parser.py
```diff
@@ -117,7 +117,8 @@ def parse_and_interpolate(
             content: str, parser=yaml.safe_load,
             conf_dir=None) -> dict:
         """Parse text content and perform variable interpolation on result."""
-        interpol_vars = parser(content).get("vars", {})
+        parsed_content = parser(content) or {}
```
IvoTod, line 120: Why are we doing this now?

IvoTod (approved):

dae/dae/configuration/gpf_config_parser.py
lchorbadjiev: If the content is an empty string (""), the `yaml.safe_load` would return `None`, breaking the downstream code.

lchorbadjiev:
merged

PR #241, opened 2022-10-12
GPF-1796 Implement logging in with a Google account

iordanivanov:
## Background

OAuth allows users to log in to third-party websites using their account on another platform, like Google.

## Aim

After implementing OAuth authentication in GPF, we wanted to allow users who are registered in the system with a Google email account to be able to login via Google instead of having to type in their password at our auth server.

## Implementation

An additional button is added to the login template, which when clicked will redirect the user to Google's authentication page. The login procedure there is up to Google, after which the user is redirected back to GPF with the appropriate auth code. We use this code to obtain an ID token from Google, see if there is a user with such an email registered in our application, and if yes, authenticate him and redirect him to where he logged in from.

### DIFF
--- a/wdae/wdae/wdae/settings.py
+++ b/wdae/wdae/wdae/settings.py
@@ -2,6 +2,7 @@
 # flake8: noqa
 
 # pylint: disable=wildcard-import,unused-wildcard-import
+import os
 from .default_settings import *
 
 INSTALLED_APPS += [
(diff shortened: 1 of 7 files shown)


wdae/wdae/wdae/settings.py
```diff
@@ -27,3 +28,9 @@
 STUDIES_EAGER_LOADING = False
 
 OPEN_REGISTRATION = True
+
+GOOGLE_AUTH_URL = os.environ["GOOGLE_AUTH_URL"]
+GOOGLE_TOKEN_URL = os.environ["GOOGLE_TOKEN_URL"]
+GOOGLE_CLIENT_ID = os.environ["GOOGLE_CLIENT_ID"]
+GOOGLE_CLIENT_SECRET = os.environ["GOOGLE_CLIENT_SECRET"]
+GOOGLE_REDIRECT_URI = os.environ["GOOGLE_REDIRECT_URI"]
```
lchorbadjiev, line 36: What happens when there is no "GOOGLE_AUTH_URL" in the environment??

lchorbadjiev (changes requested):

wdae/wdae/wdae/settings.py
lchorbadjiev: And there are no tests for this code.

PR #244, opened 2022-10-13
Fix gene score histograms

### LINKED ISSUE #240: [BUG] Gene score histograms are broken
**What happened**:

The gene score histograms in Genotype Broser are broken:

image

It is expected to see the corresponding gene score histogram.

1. Import `iossifov_2014` into `data-hg19-startup`
2. Start the `wdae`
3. Go to Genotype Browser
4. Select gene score

**Environment**:

Development version of gpf/gpfjs from the master.

iordanivanov:
Closes #240

### DIFF
--- a/wdae/wdae/gene_scores/views.py
+++ b/wdae/wdae/gene_scores/views.py
@@ -91,21 +91,24 @@ class GeneScoresPartitionsView(QueryBaseView):
 
         if not self.gpf_instance.has_gene_score(score_name):
             return Response(status=status.HTTP_404_NOT_FOUND)
+        if "min" not in data or "max" not in data:
+            return Response(status=status.HTTP_400_BAD_REQUEST)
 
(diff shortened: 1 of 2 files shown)


wdae/wdae/gene_scores/views.py
```diff
@@ -106,6 +108,7 @@ def post(self, request):
             score_max = float("inf")
```
lchorbadjiev, line 108: I would suggest using `score.min()` and `score.max()` instead of `float("inf")` when the min/max is not convertible to float. 

And there are no tests to cover this case.

lchorbadjiev (approved):
Please consider adding more tests before merging.

wdae/wdae/gene_scores/views.py
lchorbadjiev: It is OK. But add a test to cover this case.

IvoTod (approved):

PR #247, opened 2022-10-14
Management routes

IvoTod:
## Background

We are updating the frontend's management pages to be more responsive and stable. This requires a few changes in the API, as the old way of constructing the management pages required way too many queries.

## Aim

We are changing the existing users and groups API listing methods to include pagination for finer control over how many resources are loaded at once. We are also providing a new API route for dataset permissions.
Provide a short summary of the changes. Try to include answers to the **Why?** questions,
e.g. **Why are we introducing this change?**, **Why now?**, etc.

If your changes make UI changes, consider providing before and after screenshots.
You can also include a video demonstrating the old vs. the new UX.

## Implementation

The groups and users API had explicitly disabled pagination, which is now re-enabled and uses a custom pagination class.
The default page pagination class would return a response, which was way too detailed and it included things like URLs for the next and previous pages and total page count along with the body. Due to the nature of the frontend pagination mechanism, we only need to return the main body of the response and a 404 if the page is out of range. Testing fixtures for many users and groups have also been added with bulk_create

### DIFF
--- a/wdae/wdae/groups_api/tests/test_groups_rest.py
+++ b/wdae/wdae/groups_api/tests/test_groups_rest.py
@@ -378,4 +378,4 @@ def test_groups_search_pagination(admin_client, hundred_groups):
     assert len(response.data) == 25
     url = "/api/v3/groups?page=5&search=Group"
     response = admin_client.get(url)
-    assert response.status_code == status.HTTP_400_BAD_REQUEST
+    assert response.status_code == status.HTTP_404_NOT_FOUND
--- a/wdae/wdae/groups_api/tests/test_groups_rest.py
+++ b/wdae/wdae/groups_api/tests/test_groups_rest.py
@@ -1,3 +1,4 @@
+# pylint: disable=W0621,C0114,C0116,W0212,W0613
 import json
 
 from django.contrib.auth.models import Group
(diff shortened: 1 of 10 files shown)


wdae/wdae/groups_api/tests/test_groups_rest.py
```diff
@@ -329,3 +330,52 @@ def test_cant_revoke_default_permissions(user_client, dataset):
...
+    assert len(response.data) == 1
+    assert response.data[0]["name"] == "admin"
+
+def test_groups_search(admin_client, hundred_groups):
+    url = "/api/v3/groups?search=Group1"
+    response = admin_client.get(url)
+    assert len(response.data) == 12
+
+def test_groups_search_pagination(admin_client, hundred_groups):
+    url = "/api/v3/groups?page=1&search=Group"
```
svetlin-mladenov, line 367: I know I am not a reviewer but I will allow myself the luxury of sticking my nose in:
This test can be made much clearer by using parameterizing it.

wdae/wdae/users_api/tests/test_users_api.py
```diff
@@ -84,3 +85,78 @@ def test_password_validation():
     assert not is_password_valid("qwerty123456")
 
     assert is_password_valid(generic_password)
+
+
+def test_users_pagination(admin_client, hundred_users):
+    url = "/api/v3/users?page=1"
```
svetlin-mladenov, line 91: ditto

wdae/wdae/groups_api/tests/test_groups_rest.py
IvoTod: Fixed

wdae/wdae/users_api/tests/test_users_api.py
IvoTod: Fixed

lchorbadjiev (approved):

svetlin-mladenov (approved):

PR #261, opened 2022-10-25
Fix pedigrees download

### LINKED ISSUE #258: [BUG] Tagged pedigree download is slow
**What happened**:
When you use pedigree tag queries in a study's common reports, the download of selected pedigrees is painfully slow.

It should be much faster. The pedigree data is relatively small, and queries over it should work much more quickly.

Use `data-hg38-production` on Seqpipe's cluster for experimenting.

**Anything else we need to know?**:
I suspect that the slowness is caused by the usage of the `ped_df` method (property) in the `FamilyData` class. Turning families' data into a data frame is slow and should be avoided.

**Environment**:

- GPF version: development version of GPF 3.7dev from the master branch

IvoTod:
## Background

We added the functionality to download a filtered list of pedigrees by tags. The initial implementation was slow for some reason and this had to be fixed.

## Aim

To fix the slow download of pedigrees when filtering by tags.

## Implementation

The bug was due to pandas Dataframe.to_csv method returning the result as an entire string, which was later passed on to a streaming response. This resulted in a lot of overhead because every single character was being streamed individually.
I decided to change the method from using dataframes, as it could also lead to a slower download when working with larger pedigrees.
There were also 2 download routes, of which one supported tags and had a complete response with all fields, and one which just returned all pedigrees but with more limited data. The code from the latter has been cleaned up and they have been merged into one to avoid route repetition.

Closes #258

### DIFF
--- a/dae/dae/pedigrees/family.py
+++ b/dae/dae/pedigrees/family.py
@@ -345,24 +345,6 @@ class Family:
         columns.extend(sorted(extension_columns))
         return columns
 
-    def to_rows(self, sep="\t", columns=None):
-        if columns is None:
-            columns = self.get_columns()
-        rows = []
-        for member in self.full_members:
-            record = copy.deepcopy(member._attributes)
-            record["mom_id"] = member.mom_id if member.mom_id else "0"
-            record["dad_id"] = member.dad_id if member.dad_id else "0"
-            record["generated"] = member.generated \
-                if member.generated else False
-            record["not_sequenced"] = member.not_sequenced \
-                if member.not_sequenced else False
-            row = []
-            for col in columns:
-                row.append(str(record[col]))
-            rows.append(f"{sep.join(row)}\n")
-        return rows
-
     def add_members(self, persons: List[Person]) -> None:
         assert all(isinstance(p, Person) for p in persons)
         assert all(p.family_id == self.family_id for p in persons)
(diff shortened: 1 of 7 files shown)


wdae/wdae/common_reports_api/views.py
```diff
@@ -113,15 +111,14 @@ def post(self, request):
...
 
-class FamiliesDataTagDownload(QueryBaseView):
-    def get(self, request):
-        study_id = request.GET.get("study_id")
+class FamiliesDataDownloadView(QueryBaseView):
+    def get(self, request, dataset_id):
         tags = request.GET.get("tags")
 
-        if study_id is None:
+        if dataset_id is None or dataset_id == "":
```
svetlin-mladenov, line 118: I think: `if dataset:` is shorter and cleaner.

dae/dae/pedigrees/family.py
```diff
@@ -330,6 +330,39 @@ def to_json(self):
...
+            or col in ["generated", "not_sequenced"]
+        ]
+        extension_columns = column_names.difference(set(columns))
+        extension_columns = extension_columns.difference(
+            set(["sample_index"])
+        )
+        columns.extend(sorted(extension_columns))
+        return columns
+
+    def to_rows(self, sep="\t", columns=None):
```
svetlin-mladenov, line 348: I really hate the name `to_rows` - it is so general and vague. This method does something very, very specific - it serializes the object to a list of string, each string having the attributes of a member tab separated.

Personally I think you should move this method (and the to_columns method) to a seprate class, let's say `FamilyStreamTsvSerializer` that has a method serialize that returns what to_rows returns now.

lchorbadjiev (approved):

wdae/wdae/common_reports_api/views.py
IvoTod: fixed

dae/dae/pedigrees/family.py
IvoTod: fixed

wdae/wdae/common_reports_api/views.py
IvoTod: fixed

dae/dae/pedigrees/family.py
IvoTod: fixed

svetlin-mladenov (approved):

PR #262, opened 2022-10-25
Fix eager loading of studies in `wdae` and test it

### LINKED ISSUE #252: [BUG] `EAGER_LOADING` flag in Django settings does not trigger eager loading.
**What happened**:
Setting the `EAGER_LOADING` flag in Django settings to `True` does not trigger the eager loading of the GPF instance.

It is expected that when this flag is set, the GPF instance properties are eagerly loaded at the start of the application.

**Environment**:

- GPF version: development version from the master branch.

lchorbadjiev:
## Background

The `wdae` Django application assumes it works with a single `GPFInstance.`  To ensure that the `GPFInstance` object is a singleton, we create it early in the lifecycle of the Django application, using the `gpf_instance.apps.WDAEConfig` application configuration.

Depending on the Django settings flag `settings.STUDIES_EAGER_LOADING`, the initialization of the `GPFInstance` in `gpf_instance.apps.WDAEConfig` is different.

## Aim

Fixing the behavior of `gpf_instance.apps.WDAEConfig` to respect the `settings.STUDIES_EAGER_LOADING` is easy. Testing it was hard. 

The `pytest-django` plugin we use to test our Django application instantiates `gpf_instance.app.WDAEConfig` only once. I could not find a way to test different scenarios for initialization of `wdae` with different instances of `GPFInstance using `pytest-django.` 

## Implementation

To test the fix for handling `settings.STUDIES_EAGER_LOADING`, I decided to work around using `pytest-django.` 

A new package, `wdae/tests/,` is created that does not initialize the `pytest-django` plugin.

A `pytest` fixture `wdae_setup_django` that setups a Django live instance is created. This fixture works as a context manager and accepts a `GPFInstance` object and a Django settings module as arguments. The context manager starts a live Django server on a random port and returns a reference with the `url` property. This `url` could be used for running requests to the server using the `requests` library.

Closes #252

### DIFF
--- a/wdae/tests/integration/test_wdae_config/conftest.py
+++ b/wdae/tests/integration/test_wdae_config/conftest.py
@@ -12,6 +12,14 @@ class LiveServer:
     """The liveserver fixture.
 
     Copied from `pytest_django`: https://github.com/pytest-dev/pytest-django/
+
+    This class is an exact copy of the one from the `pytest-django` plugin.
+    The problem with using this class directly from the `pytest_django` package
+    is that when I import the `pytest_django.live_server_helpers.LiveServer`
+    class, it triggers the initialization of the `pytest_django` plugin,
+    which initializes our Django application. And there is no way
+    (or at least I was unable to find one) to control how the Django
+    application is initialized.
     """
 
     def __init__(self, addr: str) -> None:
--- a/wdae/tests/integration/test_wdae_config/conftest.py
+++ b/wdae/tests/integration/test_wdae_config/conftest.py
@@ -1,0 +1,155 @@
+# pylint: disable=W0621,C0114,C0116,W0212,W0613,C0415,
+# type: ignore
+
+import os
+import sys
+import contextlib
+
+import pytest
+
+
+class LiveServer:
+    """The liveserver fixture.
+
+    Copied from `pytest_django`: https://github.com/pytest-dev/pytest-django/
+    """
+
+    def __init__(self, addr: str) -> None:
+        from django.db import connections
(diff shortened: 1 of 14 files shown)


IvoTod (approved):

wdae/tests/integration/test_wdae_config/conftest.py
```diff
@@ -0,0 +1,155 @@
...
+import sys
+import contextlib
+
+import pytest
+
+
+class LiveServer:
+    """The liveserver fixture.
+
+    Copied from `pytest_django`: https://github.com/pytest-dev/pytest-django/
```
svetlin-mladenov, line 14: Can you also describe what is the difference between this class and Django's class and why you've copied it.

svetlin-mladenov (approved):

wdae/tests/integration/test_wdae_config/conftest.py
lchorbadjiev: This class is an exact copy of the one from the `pytest-django` plugin. The problem with using this class directly from the `pytest-django` package is that when I import the `pytest_django.liver_server_helpers.LiveServer` class, it triggers the initialization of the `pytest_django` plugin, which initializes our Django application. And there is no way (or at least I was unable to find one) to control how the Django application is initialized.

svetlin-mladenov: OK. Just copy paste this explanation in the docstring for the class.

lchorbadjiev: Done.

PR #274, opened 2022-11-08
GRR info index page generation

IvoTod:
## Background

We needed to generate custom HTML pages for each resource in the genomic resource repository instead of using the default file server pages.

## Aim

We are introducing this feature to make the GRR more friendly to navigate and preview info more easily.

## Implementation

2 new commands have been added to `grr_manage`. These are `repo-info` and `resource-info`. Both commands will always generate the index files. We decided to do this as it is a very light task.
The `GenomicResource` class used to be part of the implementation classes of the resources (e.g. `PositionScore`, `ReferenceGenome`, etc.). Due to problems with this design `GenomicResource` was made separate and was concerned only about repository operations. This was a problem again, as all implementation classes needed common methods to get templates for their index pages and the information to supply them. To solve this we added `GenomicResourceImplementation`.
When adding `GenomicResourceImplementation`  as a base class to the implementations, gene sets and gene scores were not straight forward and required a quick refactor to be compatible and in line with how the other classes are implemented.


dae/dae/genomic_resources/templates.py
```diff
@@ -0,0 +1,108 @@
+import math
```
lchorbadjiev, line 1: I think this should be part of the cli.py module.

lchorbadjiev (approved):
I have some minor suggestions for changes. Please address them before merging.

PR #291, opened 2022-11-15
Fix datasets duplicates

IvoTod:
## Background

In the datasets hierarchy route, studies on the top level were being duplicated and displayed as datasets who have themselves as the single child.

## Aim

To fix the duplication bug on the backend.

## Implementation

The bug was due to a difference between the interfaces of GenotypeData and StudyWrapper. In GenotypeData, the is_group method is a property, but in StudyWrapper was a property and a method depending on the case. In the route, is_group was used as a property, but because it isn't and a function reference is truthy, `is_group` was always True. To fix this, the study wrapper classes were made to mirror the genotype data classes and use properties. Fixing this behavior I noticed an erroneous  is_group property for remote genotype data. All genotype data has a list of studies internally, studies have a list of length 1 that contains themselves, so using the length of _study_ids could lead to a false positive.


wdae/wdae/studies/study_wrapper.py
```diff
@@ -231,8 +231,9 @@ def __init__(self, genotype_data_study, pheno_db, gene_scores_db):
     def __getattr__(self, name):
         return getattr(self.genotype_data_study, name)
 
+    @property
     def is_group(self):
-        return self.genotype_data_study.is_group()
+        return self.genotype_data_study.is_group
```
lchorbadjiev, line 236: It seems like we do not have tests for this method of StudyWrapper. Could you please add some?

wdae/wdae/studies/remote_study.py
```diff
@@ -93,7 +93,7 @@ def description(self):
 
     @property
     def is_group(self):
```
lchorbadjiev, line 95: Please add tests for this method.

lchorbadjiev (changes requested):
I would like to have some tests for the methods we change.

wdae/wdae/studies/remote_study.py
IvoTod: Adding the tests proved troublesome because of remote dataset support, so now remote genotype datas cannot be groups.

wdae/wdae/studies/study_wrapper.py
IvoTod: Added tests

lchorbadjiev (approved):

PR #302, opened 2022-11-22
Continue support for import tools

svetlin-mladenov:
## Background

The import process is a slow one. If something breaks we would like to have a mechanism to continue from where we left off.

## Aim

For each task in a task graph after its completion we store the result on disk. Next time we run import_tools if the result files are present and there haven't been any changes to the input files then we can safely load the result from disk and skip recalculation.

## Implementation

Task Graph related code moved to it's own module `dae.task_graph`. Added a new class TaskCache which takes care of storing and loading task results from disk. Currently pickle is used to store the results.

The import tools cli has been changed a little bit. It now accepts the import configuration as a (first) position argument. No need for `-f`. `--force` (or `-f` for short) is now used to ignore stored results and force recalculation of the entire graph.


dae/dae/task_graph/executor.py
```diff
@@ -0,0 +1,239 @@
...
+            record = self._task_cache.get_record(task_node)
+            if record.type == CacheRecordType.COMPUTED:
+                already_computed_tasks[task_node] = record.result
+                self._set_task_result(task_node, record.result)
+            else:
+                self._queue_task(task_node)
+        return self.__await_tasks(already_computed_tasks)
+
+    def __await_tasks(self, already_computed_tasks) \
+            -> Iterator[tuple[Task, Any]]:
```
lchorbadjiev, line 55: This name is too similar to `_await_tasks,` and it was confusing at first.

lchorbadjiev (approved):

dae/dae/task_graph/executor.py
svetlin-mladenov: @lchorbadjiev Fixed.

Issue #280, opened 2022-11-09
[BUG] Unstable test - users_api.tests.test_users_authentication.test_failed_auth_lockouts


iordanivanov:
Stack trace:
```python
db = None, user = <WdaeUser: user@example.com>
client = <django.test.client.Client object at 0x7f2b6cb2abb0>
tokens = (<AccessToken: user-token>, <AccessToken: admin-token>)

    def test_failed_auth_lockouts(db, user, client, tokens):
        # Check if progressive lockouts are working.
        url = "/api/v3/users/login"
        data = {
            "username": "user@example.com",
            "password": "wrongpassword",
        }
    
        lockout_email(client, data["username"])
    
        for i in range(1, 5):
            response = client.post(
                url, json.dumps(data),
                content_type="application/json", format="json"
            )
            assert response.content.find(b"This account is locked out for") != -1
            regex = r"locked out for (\d+) hours and (\d+) minutes"
            match = re.search(regex, response.content.decode())
            response_hours, response_minutes = map(int, match.groups())
            total_minutes = pow(2, i)
            hours = int(total_minutes / 60)
            minutes = total_minutes % 60
    
            assert response_hours == hours
>           assert response_minutes == minutes
E           assert 1 == 2

wdae/users_api/tests/test_users_authentication.py:189: AssertionError
```

IvoTod:
Fixed in #315

PR #315, opened 2022-11-29
Stabilize flaky users_api authentication tests

### LINKED ISSUE #279: [BUG] Unstable test - users_api.tests.test_users_authentication.test_authentication_logging
This unit test can randomly fail and break builds. Fix it to be consistent.

iordanivanov:
Should fix #279 and #280.

## Background

Two of the users_api authentication tests would fail randomly, breaking the Jenkins build. They are related to timestamps for logging in and locking out users and were written without allowing a tolerance between the received and expected timestamps, which would make these tests fail at random.

## Aim

Stabilize the tests.

## Implementation

For the lockout test, I used pytest's `approx` functionality and gave a tolerance of one minute. For the login timestamp tests, since they use Django timestamp objects, I ignore the `second` and `microsecond` values by replacing them with `0` in both the result and the expected value.

Closes #279 and #280.


wdae/wdae/users_api/tests/test_users_authentication.py
```diff
@@ -181,12 +182,13 @@ def test_failed_auth_lockouts(db, user, client, tokens):
...
-        hours = int(total_minutes / 60)
-        minutes = total_minutes % 60
+        response_td = timedelta(
+            hours=response_hours, minutes=response_minutes
+        )
+        expected_td = timedelta(minutes=pow(2, i))
 
-        assert response_hours == hours
-        assert response_minutes == minutes
+        # Give a tolerance of one minute to prevent test from becoming flaky
```
IvoTod, line 190: Was the test flaky enough to demand approximation up to a whole minute?

iordanivanov: I figure minute precision is good enough for testing - didn't want to experiment with 2 or 5 second tolerances

IvoTod: But if we see a difference that is almost a minute wouldn't that mean that there is an issue with the code? I think 5 seconds should be more than enough

iordanivanov: Alright - I'll change the tolerances to 5 seconds.

IvoTod (approved):

PR #316, opened 2022-11-30
Resource implementation plugins

### LINKED ISSUE #309: [ENH] Resource implementation plugins
Before, genomic resources were coupled with their implementations and adding a new resource meant creating it as a separate plugin and registering it. This was removed when implementations were separated from the genomic resources.

Now that we have a clear idea of what a genomic resource implementation is, we should reimplement them as plugins.

IvoTod:
## Background

Genomic resources used to be plugins when they were merged in one class with implementations. Ever since their split, the plugins were removed and left for later.

## Aim

Restore registering of implementations as entry points and accessing them through a central registry.

## Implementation

Every resource is expected to have a builder function, which takes a resource. This function will be registered to the resource's type, as specified in setup.py. Due to cyclic import issues, registered implementations are first discovered when running, but are loaded on demand.

Closes #309

### DIFF
--- a/dae/dae/genomic_resources/__init__.py
+++ b/dae/dae/genomic_resources/__init__.py
@@ -23,11 +23,23 @@ _PLUGINS_LOADED = False
 def get_resource_implementation_factory(
     implementation_type: str
 ) -> Callable[[GenomicResource], GenomicResourceImplementation]:
-    if implementation_type not in _REGISTERED_RESOURCE_IMPLEMENTATIONS:
-        raise ValueError(
-            f"unsupported resource implementation type: {implementation_type}"
-        )
-    return _REGISTERED_RESOURCE_IMPLEMENTATIONS[implementation_type]
+    """
+    Return an implementation builder for a certain resource type.
+
+    If the builder is not registered, then it will search for an entry point
+    in the found implementations list. If an entry point is found, it will be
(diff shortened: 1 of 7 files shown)


dae/dae/genomic_resources/__init__.py
```diff
@@ -1,24 +1,88 @@
...
+    """
+    if resource_type not in _REGISTERED_RESOURCE_IMPLEMENTATIONS:
+        if resource_type not in _FOUND_RESOURCE_IMPLEMENTATIONS:
+            raise ValueError(
+                f"unsupported resource implementation type: {resource_type}"
+            )
+        entry_point = _FOUND_RESOURCE_IMPLEMENTATIONS[resource_type]
+        loaded = entry_point.load()
+        register_implementation(resource_type, loaded)
+        _REGISTERED_RESOURCE_IMPLEMENTATIONS[resource_type] = loaded
```
iordanivanov, line 44: I may be misreading this, but isn't this line a repeat of what is done in `register_implementation`?

IvoTod: whoops

IvoTod: Fixed

iordanivanov (approved):

PR #330, opened 2022-12-02
In memory tags fix

### LINKED ISSUE #307: [BUG] Families from In-Memory genotype storage are without tags
**What happened**:

When loading families from in-memory genotype storage, they have no pedigree tags. This breaks functionality for filtering families by tags in the datasets statistics UI.

The pedigree tags should be built when loading families from in-memory genotype storage.

You can use the `helloworld` de Novo example from the getting started guide:
https://www.iossifovlab.com/gpfuserdocs/administration/getting_started.html#example-import-of-de-novo-variants-helloworld

**Environment**:

- GPF version: development version 3.7dev from the master branch

IvoTod:
## Background

Following the getting started guide and creating an in memory study resulted in no tags being generated and later no tags being loaded.

## Aim

Restore tags when following the steps in the getting started guide.

## Implementation

While the original issue expected tags to be created when loading in-memory studies, I decided to fix it by restoring the default tag generation. Before we switched to import projects, ped-tags used to be True by default, which led to all newly imported studies having tags unless explicitly specified not to.
I also fixed a bug in the base CLILoader class where cli_defaults added positional arguments to the output dictionary as the following key-value pair (None, None).

Closes #307.

### DIFF
--- a/dae/dae/import_tools/import_tools.py
+++ b/dae/dae/import_tools/import_tools.py
@@ -122,12 +122,13 @@ class ImportProject():
         """Get params for loading the pedigree."""
         families_filename = self.get_pedigree_filename()
 
-        default_families_params = FamiliesLoader.cli_defaults()
-        families_params = self.import_config["input"]["pedigree"]
-        families_params = self._add_loader_prefix(families_params, "ped_")
-        default_families_params.update(families_params)
+        families_params = {}
+        families_params.update(FamiliesLoader.cli_defaults())
+        config_params = self.import_config["input"]["pedigree"]
+        config_params = self._add_loader_prefix(config_params, "ped_")
+        families_params.update(config_params)
 
-        return families_filename, default_families_params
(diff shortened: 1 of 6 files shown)


dae/dae/import_tools/import_tools.py
```diff
@@ -122,10 +122,12 @@ def get_pedigree_params(self) -> Tuple[str, Dict[str, Any]]:
         """Get params for loading the pedigree."""
         families_filename = self.get_pedigree_filename()
 
+        default_families_params = FamiliesLoader.cli_defaults()
         families_params = self.import_config["input"]["pedigree"]
         families_params = self._add_loader_prefix(families_params, "ped_")
+        default_families_params.update(families_params)
 
-        return families_filename, families_params
+        return families_filename, default_families_params
```
lchorbadjiev, line 130: The `default_families_params` name for the returned value is confusing. At least, I was confused.

PR #339, opened 2022-12-06
Management adjustments

### LINKED ISSUE #336: [ENH] Groups API cannot retrieve group by name.
When a group's datasets or users are modified, the frontend should be able to update the affected group by retrieving it again. However, this is not possible because the DRF default behavior for viewsets uses the PK for retrieving.

We should be able to retrieve a group's info by querying `/api/v3/groups/(group_name)`

### LINKED ISSUE #337: [ENH] Dataset permissions for singular dataset
We have a route on `/api/v3/datasets/permissions` that returns the permissions info about WDAE datasets and supports searching. This does not cover the use case of fetching the permissions info about a single WDAE dataset. 

Add a route on `/api/v3/datasets/permissions/(dataset_id)` for single dataset requests.

IvoTod:
## Background

We experienced a miscommunication on what some routes should return and how the API should be used on the frontend with the new management rework. After updating a group, user or a dataset, we could only re-retrieve the user, but not the group or dataset. The groups API supports retrieving, but it requires the group ID, which is not exposed anywhere in the API.

## Aim

Add retrieve support for datasets and groups.

## Implementation

Groups API was changed to use the group name for retrieval, datasets had a new route implemented that allows a client to retrieve the dataset permissions info for a specific dataset.

Closes #336 
Closes #337

### DIFF
--- a/wdae/wdae/datasets_api/views.py
+++ b/wdae/wdae/datasets_api/views.py
@@ -268,28 +268,8 @@ class DatasetDescriptionView(QueryBaseView):
         return Response(status=status.HTTP_200_OK)
 
 
-class DatasetPermissionsView(QueryBaseView):
-
-    page_size = settings.REST_FRAMEWORK["PAGE_SIZE"]
-
-    def get(self, request):
-        dataset_search = request.GET.get("search")
-        page = request.GET.get("page", 1)
-        query = Dataset.objects
-        if dataset_search is not None and dataset_search != "":
-            query = query.filter(dataset_id__icontains=dataset_search)
-
-        if page is None:
-            return Response(status.HTTP_400_BAD_REQUEST)
-        if isinstance(page, str):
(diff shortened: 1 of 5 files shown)


wdae/wdae/datasets_api/views.py
```diff
@@ -313,33 +293,71 @@ def get(self, request):
...
+            query = query.filter(dataset_id__icontains=dataset_search)
+
+        if page is None:
+            return Response(status.HTTP_400_BAD_REQUEST)
+        if isinstance(page, str):
+            page = int(page)
+
+        page_start = (page - 1) * self.page_size
+        page_end = page * self.page_size
+        datasets = query.all()[page_start:page_end]
```
lchorbadjiev, line 334: Is this the recommended way to do paging?

lchorbadjiev (approved):

PR #343, opened 2022-12-14
Variant reports download fix

livovachkov:
## Background

Family counters download follows the old type of requests that parse JSON which is unnecessary and outdated.

## Aim

The aim is to follow the consistency of the other download requests. This is also backed up by a small change in the gpfjs method of making requests.

## Implementation

The implementation is to remove the unneeded JSON logic, change the test in order to follow the new logic, and also a new call has been added in the reset_dev.sh script in case the setup doesn't call wdae_create_dev_federation_app, which prevents testing


IvoTod (approved):
I am approving these changes, but there is 1 more test to update before merging in test_common_reports_api.py - one of the parameterized tests for permissions at the start of the file.

PR #357, opened 2023-01-09
Kubernetes bug fixes

svetlin-mladenov:
## Background

Minor bug fixes to our integration with kubernetes.

## Implementation

The only major change is making the `_gpf_instance` in `ImportProject` transient to pickle. The gpf instance contains references to a whole bunch of objects some of which may not be pickleable. We tried making these objects pickleable but that required a lot of changes and there was always some edge-case we didn't cover. Making the gpf instance transient allows us to pickle the project without worrying about the gpf instance. The gpf instance will be automatically recreated when the import project is deserialized.

### DIFF
--- a/dae/dae/dask/client_factory.py
+++ b/dae/dae/dask/client_factory.py
@@ -71,12 +71,10 @@ class DaskClient:
                 "dashboard_address":
                 f":{kwargs.get('dashboard_port', 8787)}"
             }
-        if kwargs.get("log_dir"):
-            log_dir = kwargs.get("log_dir")
-        else:
-            # pylint: disable=consider-using-with
-            tmp_dir = tempfile.TemporaryDirectory()
-            log_dir = tmp_dir.name
+
+        # pylint: disable=consider-using-with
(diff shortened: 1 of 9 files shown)


dae/dae/dask/client_factory.py
```diff
@@ -32,8 +33,8 @@ def add_arguments(parser):
             "kubernetes workers")
         group.add_argument(
             "--container-image",
-            default="registry.seqpipe.org/seqpipe-gpf:"
-            "dask-for-hist-compute_fc69179-14",
+            default="registry.seqpipe.org/iossifovlab-gpf:"
+            "import-k8s-support_bee6c29-7",
```
lchorbadjiev, line 37: This default value is not good. It should be changed before merging.

lchorbadjiev (approved):

dae/dae/dask/client_factory.py
svetlin-mladenov: Fixed!

PR #360, opened 2023-01-11
Add file locking when caching filesystem genomic resource

### LINKED ISSUE #335: Lock genomic resource's files while caching

iordanivanov:
## Background

When caching a resource, problems would occur when another process would try to cache the same resource, leading to the resource's files being overwritten with invalid data. 

## Aim

Lock genomic resources' files if they are currently being written to, and release the lock when the file has been cached.

## Implementation

The current implementation has been done only for filesystem repositories, S3 repositories do not have locking. Locking has been implemented by creating a .lockfile for the genomic resource file being cached, and placing a lock on it using the `fcntl` module. The lock is obtained in CachingProtocol before the `update_resource_file` method is called, and automatically released after that method finishes.

## Related issues

Closes #335

### DIFF
--- a/dae/dae/genomic_resources/fsspec_protocol.py
+++ b/dae/dae/genomic_resources/fsspec_protocol.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import os
+import fcntl
 import hashlib
 import logging
 import datetime
@@ -159,6 +160,7 @@ class FsspecReadOnlyProtocol(ReadOnlyRepositoryProtocol):
         compression = None
         if kwargs.get("compression"):
             compression = "gzip"
+
         return self.filesystem.open(
             filepath, mode=mode,
             compression=compression)  # pylint: disable=unspecified-encoding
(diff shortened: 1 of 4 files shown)


dae/dae/genomic_resources/tests/test_cached_repo.py
```diff
@@ -238,3 +243,79 @@ def test_cached_repository_file_level_cache(cache_repository, scheme):
...
+            }}, scheme=scheme) as cache_repo:
+
+        resource = cache_repo.get_resource("one")
+        assert resource is not None
+
+        orig = resource.proto.local_protocol.copy_resource_file
+
+        def block(*args):
+            time.sleep(5)
+            orig(*args)
```
lchorbadjiev, line 302: I don't like to have `time.sleep` in tests. I would prefer to get rid of it.

dae/dae/genomic_resources/tests/test_cached_repo.py
```diff
@@ -238,3 +243,79 @@ def test_cached_repository_file_level_cache(cache_repository, scheme):
...
+                                 args=("data.txt",),
+                                 daemon=True)
+            x.start()
+            y = threading.Thread(target=resource.open_raw_file,
+                                 args=("data.txt",),
+                                 daemon=True)
+            y.start()
+            time.sleep(0.25)  # for test stability
+            assert protocol_spy.call_count == 1
+            time.sleep(5)
```
lchorbadjiev, line 320: We need to get rid of these `time.sleep` calls.

dae/dae/genomic_resources/fsspec_protocol.py
```diff
@@ -219,6 +221,42 @@ def __init__(
...
+                self.filesystem.makedirs(
+                    os.path.dirname(path), exist_ok=True)
+            lockfile = open(path, "wt", encoding="utf8")
+            lockfile.write(str(datetime.datetime.now()) + "\n")
+            fcntl.flock(lockfile, fcntl.LOCK_EX)
+            lock.__enter__ = lockfile.__enter__
+            lock.__exit__ = lockfile.__exit__
+
+        return lock
+
```
lchorbadjiev, line 259: I am not sure, but it seems to me that it would be better to raise `NotImplementedError` if we do not support locking. Probably it needs discussion.

lchorbadjiev (changes requested):
We need to discuss what to do if locking is not supported and how to get rid of calls to `time.sleep`.

lchorbadjiev (approved):

PR #370, opened 2023-01-27
Resource statistics framework

IvoTod:
## Background

As we added more and more implementations, we needed a uniform framework for generating and using additional statistics on each implementation.

## Aim

Create a framework for implementing, generating and using resource statistics with task graphs.

## Implementation

Resource statistics replace histograms in the GRR CLI. The base implementation class has been reworked to be more minimal and now requires each implementation to provide task graph tasks for execution for statistics generation. The GRR CLI commands have been reworked to depend on one another following a hierarchy. The new stats command collects and executes tasks for each resource. For the initial step, genomic score statistics have been implemented.
There are a few things on which I am not sure and should be considered:
 - Where should we place our statistics code? There are quite a few planned so in a single file is probably not the best option.
 - How should we change the DaskClient + DaskExecutor and SequentialExecutor so they can be more compatible when used in an if branch together.
 - Running the statistics command on a repository can result in a segmentation fault only sometimes before it starts executing tasks.


dae/dae/genomic_resources/tests/test_score_statistics_state.py
```diff
@@ -54,105 +53,105 @@ def proto_fixture(tmp_path):
     return build_filesystem_test_protocol(tmp_path)
 
 
-def test_double_build_histograms_without_change(proto_fixture, dask_client):
+# def test_double_build_histograms_without_change(proto_fixture, dask_client):
```
svetlin-mladenov, line 56: If that test doesn't work anymore either fix it or remove it.

dae/dae/genomic_resources/tests/test_cli_hist.py
```diff
@@ -1,209 +1,209 @@
...
-
-    # Then
-    out, err = capsys.readouterr()
-    assert out == ""
-    assert err == ""
-    assert caplog.record_tuples == [
-        ("dae.genomic_resources.histogram", logging.INFO,
-         "histograms of <one> are up to date")
-    ]
+# import os
```
svetlin-mladenov, line 2: Delete these tests if they are not relevant anymore.

dae/dae/genomic_resources/repository.py
```diff
@@ -547,6 +547,8 @@ def check_update_manifest(
         manifest = Manifest()
         entries_to_update = set()
         for entry in self.collect_resource_entries(resource):
+            if entry.name.endswith("html"):  # Ignore generated info files
```
svetlin-mladenov, line 550: I don't understand that if

dae/dae/genomic_resources/cli.py
```diff
@@ -361,144 +383,185 @@ def _run_resource_manifest_command(proto, repo_url, **kwargs):
...
-        client, proto, res, dry_run, force, use_dvc, region_size):
-    if res.get_type() not in {
-            "position_score", "np_score", "allele_score"}:
+
+def _collect_impl_stats_tasks(  # pylint: disable=too-many-arguments
+        graph, proto, impl, dry_run, force, use_dvc, region_size):
+
+    impl.add_statistics_build_tasks(graph, region_size=region_size)
+
+    stats_tasks = copy.copy(graph.tasks)
```
svetlin-mladenov, line 416: No need to do that copy

dae/dae/genomic_resources/repository.py
IvoTod: It was decided to ignore generated info pages in the manifest as their generation is fast and they are lightweight. If info pages are not ignored, then the manifest would also become immediately outdated the moment we run the info command.

dae/dae/genomic_resources/tests/test_cli_hist.py
IvoTod: Done

dae/dae/genomic_resources/tests/test_score_statistics_state.py
IvoTod: Deleted.

dae/dae/genomic_resources/cli.py
IvoTod: Fixed

svetlin-mladenov (approved):
I skimmed the changes quickly. They look OK to me.

PR #373, opened 2023-01-30
Command line interface helpers for executing task graphs.

lchorbadjiev:
## Background

We have a task graph abstraction and executors that execute tasks from these task graphs. We want to use these abstractions in various command line tools - import tools, GRR manage, etc.

We need a common infrastructure for executing tasks from task graphs. 

This work was started by @iossifov and was integrated by @svetlin-mladenov.

## Aim

To have a set of helper functions that allow embedding task graph execution in command line tools.

## Implementation

There is a `cli_tools.TaskGraphCli` class that can:
- add parameters to existing `argparse` argument parser
- create an executor from the parameters collected from the argument parser
- process a task graph in accordance with the parameters collected from the argument parser

Helper functions to execute `run` and `status` commands are added to `dae.task_graph.executor` module.

**Why it's implemented the way it is? What alternative implementations have you
considered and why you chose this one?**

The `dae.dask.client_factory` module is deprecated and should be removed soon.

### DIFF
--- a/dae/dae/dask/named_cluster.yaml
+++ b/dae/dae/dask/named_cluster.yaml
@@ -39,13 +39,12 @@ dae_named_cluster:
     - name: kubernetes
       type: kubgernetes
       params:
-        container_image: xxxxx
-        extra_pod_config:
-          A: B
-          b: 234234
-        envvars: 
-          - PATH
-          - MY_ENV
+        container_image: registry.seqpipe.org/iossifovlab-gpf:latest
+        # extra_pod_config:
+        #   A: B
+        #   b: 234234
+        # envvars: 
+        #   - PATH
     - name: slurm
       type: slurm
       number_of_threads: 10
--- a/dae/dae/dask/named_cluster.yaml
+++ b/dae/dae/dask/named_cluster.yaml
@@ -1,5 +1,5 @@
 dae_named_cluster:
   default: default
   clusters:
-    default: 
+    - name: default 
       type: local
--- a/dae/dae/dask/named_cluster.yaml
+++ b/dae/dae/dask/named_cluster.yaml
@@ -1,18 +1,41 @@
 dae_named_cluster:
   default: local
   clusters:
-    - name: local 
+    - name: local
       type: local
-    - name: sge_quick 
+      # By default, the local cluster will allocate by default as many threads 
+      # as the number of cores on the host, spread accross several workers each
+      # with cerain number of threads.  See the documentation of LocalCluster
+      # for more information. 
+      # To set the default a number of threads we can set:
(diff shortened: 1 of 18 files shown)


dae/dae/dask/named_cluster.yaml
```diff
@@ -0,0 +1,57 @@
...
+      params:
+        scheduler_options:
+          dashboard_address: :8898
+        cores: 1
+        memory: 10GB
+        log_directory: ./sge_worker_logs
+    - name: kubernetes
+      type: kubgernetes
+      params:
+        container_image: xxxxx
```
svetlin-mladenov, line 42: These are the default values, right? Let's put something more appropriate here like `registry.seqpipe.org/iossifovlab-gpf:latest`

dae/dae/dask/named_cluster.yaml
```diff
@@ -0,0 +1,57 @@
...
+        extra_pod_config:
+          A: B
+          b: 234234
```
svetlin-mladenov, line 45: Maybe remove these example values

dae/dae/task_graph/__init__.py
```diff
@@ -0,0 +1,4 @@
+# flake8: noqa: F401
+from .cli_tools import TaskGraphCli
+from .graph import TaskGraph
+from .graph import Task
```
svetlin-mladenov, line 4: I think, in order to be consistent, we should use absolute imports here.

dae/setup.py
```diff
@@ -96,6 +96,7 @@
     stats_liftover=dae.tools.stats_liftover:main
 
     import_tools=dae.import_tools.cli:main
+    task_graph_demo=dae.task_graph.demo_graphs_cli:main
```
svetlin-mladenov, line 99: Do we really want that? Do we really want people when they install gpf to get a `task_graph_demo` executable on their path? A demo that doesn't demo anything client related.

svetlin-mladenov (approved):

dae/dae/dask/named_cluster.yaml
lchorbadjiev: fixed

dae/dae/dask/named_cluster.yaml
lchorbadjiev: I commented them. But want to keep them as an example.

dae/dae/task_graph/__init__.py
lchorbadjiev: fixed

dae/setup.py
lchorbadjiev: fixed

PR #375, opened 2023-01-31
Implement per-file caching granularity for use in grr_cache_repo tool

### LINKED ISSUE #364: [ENH] Selectively cache resource files with grr_cache_repo tool
The `grr_cache_repo` tool can cache specific resources from a repository, but it will cache the entire resource which is undesirable if we only need some specific files from it (for example when annotating).

Allow the grr_cache_repo tool to cache only specific files from a given resource. Implement functionality in annotator classes that will provide which files this annotator will need from its given resource. Collect a list of all files from all annotators and feed it into the `grr_cache_repo` tool.

iordanivanov:
## Background

The grr_cache_repo tool exists in order to cache a number of resources according to a given configuration. It will cache all resources that are used by that configuration, but it will cache the entire resource instead of just the files needed from the resource. This can be wasteful with huge resources where we don't always need every single file.

## Aim

When caching for a provided annotation config, make it so that the tool will cache only the files that it will use, instead of the complete contents of each resource.

## Implementation

Functionality was implemented in GRR protocols to allow passing a list of which files to cache from a resource, instead of iterating the manifest and caching every file found in it. If that list is not given, the old greedy behavior is used. Each annotator and their related resource implementation have been modified so that they can provide a list of files that they actually use from whatever resource they are based on. Then, the grr_cache_repo tool will collect all of these files into a "resource -> file list" dictionary and pass it down to the caching protocol.

## Related issues

Closes #364


dae/dae/genomic_resources/repository.py
```diff
@@ -738,21 +738,34 @@ def copy_resource(
             version_constraint=f"={remote_resource.get_version_str()}")
 
     def update_resource(
```
lchorbadjiev, line 740: It seems to me that this should be renamed to something like `update_resource_files`.

The behavior of this method when the set of files is `None` should be described in the doc string.

dae/dae/genomic_resources/repository.py
```diff
@@ -738,21 +738,34 @@ def copy_resource(
...
-            self, remote_resource: GenomicResource) -> GenomicResource:
-        """Copy a remote resource into repository."""
+        self,
+        remote_resource: GenomicResource,
+        files: Optional[Set[str]] = None
+    ) -> GenomicResource:
+        """Copy a remote resource into repository.
+
+        Allows copying of only specific files from the resource.
+        If no files are explicitly given, copies the whole resource.
```
lchorbadjiev, line 748: It seems to me that it copies only the resource configuration file?

lchorbadjiev (changes requested):
The implementation of the `resource_files` property is kind of not coherent. A couple of resources have the `files` property, which makes the `resource_files` implementation very straightforward. Other resources do not have the `files` property, and we need to parse the resource configuration in the annotator, which looks ugly.

It is not obvious what the `ReadWriteRepositoryProtocol.update_resource` method does. Let us talk about how it works.

lchorbadjiev (approved):

PR #382, opened 2023-02-08
Genomic position table improvements

### LINKED ISSUE #345: Make VCF genome position table score configs "merge" with autodetected configs
So that we can specify a number of scores from the INFO field we wish to use, but also make use of the information that is provided by the VCF header such as type, number, etc.

### LINKED ISSUE #374: [BUG] Genomic position table does not handle non-attribute columns with identical indices properly
**What happened**:
Given the sample configuration below, when accessing index number 4 and 5, we receive the alternative allele and the first score value.

When accessing index number 4 and 5, we expect the two score values.

```python
def test_overlapping_nonattribute_columns_config(tmp_path):
    setup_directories(
        tmp_path, {
            "genomic_resource.yaml": """
                table:
                  filename: data.txt.gz
                  format: tabix
                  header_mode: none
                  chrom:
                    index: 0
                  pos_begin:
                    index: 1
                  pos_end:
                    index: 1
                  reference:
                    index: 2
                  alternative:
                    index: 3
                scores:
                  - id: raw
                    index: 4
                    type: float
                    desc: "raw"
                  - id: phred
                    index: 5
                    type: float
                    desc: "phred"
            """,
        })
    setup_tabix(
        tmp_path / "data.txt.gz",
        """
        1       101   A       C       0.123        1
        1       102   A       C       0.456        2
        1       103   A       C       0.789        3
        """, seq_col=0, start_col=1, end_col=1)
    res = build_filesystem_test_resource(tmp_path)
    with build_genomic_position_table(res, res.config["table"]) as tab:
        results = tuple(map(lambda l: (l[4], l[5]), tab.get_all_records()))
        assert results == (
            ("0.123", "1"),
            ("0.456", "2"),
            ("0.789", "3"),
        )
```

iordanivanov:
## Background

Two issues are solved in this PR:
- We would like to be able to select a subset of auto-configured scores in VCF genomic position tables
- Handling of a single position column in genomic position tables is broken because it does not respect original indices, offsetting them by one (by creating a pos end column)

## Implementation

- If a VCF genomic position table is given some score columns in its configuration, it will ignore any columns that are not specified in the config and will merge autogenerated values with those given in the configuration (type, description, etc.)
- Line adapters (Line and VCFLine classes) have been refactored to work in a simpler manner, where they now only provide access to a number of core attributes (such as `chrom`) and a `get` method that works with an index (`int`) or a column name (`str`) if a header was provided when the adapter was instantiated. All behaviour that imitates tuples has been removed.

## Related issues

Closes #345
Closes #374


dae/dae/genomic_resources/genomic_position_table/line.py
```diff
@@ -47,15 +49,22 @@ def __iter__(self):
...
-        if isinstance(key, slice):
-            return tuple(self)[key]
-        if key == 0:
-            return self.chrom
-        if key == 1:
-            return self.pos_begin
-        if key == 2:
-            return self.pos_end
-        return tuple(self.attributes.values())[key - 3]
+        if self.single_position:
```
lchorbadjiev, line 52: I wouldn't say I like this fix. Let us discuss this.

dae/dae/genomic_resources/genomic_position_table/table_inmemory.py
```diff
@@ -71,6 +71,7 @@ def open(self):
             chrom = columns[self.chrom_column_i]
             ps_begin = int(columns[self.pos_begin_column_i])
             ps_end = int(columns[self.pos_end_column_i])
+            single_pos = self.pos_begin_column_i == self.pos_end_column_i
```
lchorbadjiev, line 74: This should be calculated once when the table is configured or opened.

dae/dae/genomic_resources/genomic_position_table/table_tabix.py
```diff
@@ -245,11 +245,13 @@ def get_line_iterator(self, chrom=None, pos_begin=None):
                                              self.pos_end_column_i)}
             ref = attributes.get(self.ref_key)
             alt = attributes.get(self.alt_key)
+            single_pos = self.pos_begin_column_i == self.pos_end_column_i
```
lchorbadjiev, line 248: This should be calculated once when the table is configured or opened.

lchorbadjiev (changes requested):
I want to discuss how to handle getting values from a line. 

I tend to prefer having two classes - `PositionLine` and `RegionLine` to handle the difference in `__getitem__` method.

lchorbadjiev (approved):

PR #392, opened 2023-02-24
Handling for chromosome only region queries #707

livovachkov:
## Background

Currently, the region filter requires positions in order to get the variants between them. There is no support for region filtering on a whole chromosome.

## Aim

Provide support for handling chromosome filtering in region filter.

## Implementation

Linked with front-end issue #707


lchorbadjiev (changes requested):
Add tests for these changes.

PR #408, opened 2023-03-14
Reorganization of `wdae.default_settings`

lchorbadjiev:
## Background

Currently, the `wdae` application has too many different `wdae.settings` that could come from different places. 

## Aim

We want to collect all important configuration variables for the project in `wdae.default_settings` and make them configurable through environment variables. The `wdae.settings` in `iossifovlab-gpf-containers/iossifovlab-gpf-full` should become trivial and should not contain important overrides.

## Implementation

We moved most of the configuration parameters in `wdae.default_settings` and made them configurable through environment variables. In the setup, we run `MailHog` to make possible integration testing for sending mails. Example reset password test is added.

### DIFF
--- a/wdae/wdae/wdae/default_settings.py
+++ b/wdae/wdae/wdae/default_settings.py
@@ -5,7 +5,13 @@ import os
 import logging
 from dae.pheno.pheno_db import get_pheno_browser_images_dir
 
-DEBUG = True
+DEBUG = os.environ.get("WDAE_DEBUG", "False") == "True"
+
+
+# Make this unique, and don't share it with anybody.
+SECRET_KEY = os.environ.get(
+    "WDAE_SECRET_KEY", "#mhbhbjgub==v$cjdu7jay@*$ux$novw#t2tmgjr^5(pr@ycxp"
+)
 
 
 STUDIES_EAGER_LOADING = False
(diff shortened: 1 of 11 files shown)


wdae/wdae/wdae/default_settings.py
```diff
@@ -28,41 +34,73 @@
...
+
+if WDAE_PREFIX:
+    LOGIN_URL = f"/{WDAE_PREFIX}/accounts/login/"
+    FORCE_SCRIPT_NAME = f"/{WDAE_PREFIX}"
+else:
+    LOGIN_URL = "/accounts/login/"
+    FORCE_SCRIPT_NAME = "/"
+
+EMAIL_HOST = os.environ.get("WDAE_EMAIL_HOST", "localhost")
+EMAIL_USET_TLS = os.environ.get("WDAE_EMAIL_USET_TLS", False)
```
IvoTod, line 50: Shouldn't this be EMAIL_USE_TLS?

IvoTod (approved):

wdae/wdae/wdae/default_settings.py
lchorbadjiev: fixed

PR #409, opened 2023-03-16
Caching status update

### LINKED ISSUE #400: [ENH] GRR resource file caching status indicator
When files from GRR resources are being cached, there is no status indicator.

A status indicator that outputs which resource and which file from that resource is are currently being cached. If possible, also display how many files there are in total to be cached.

livovachkov:
## Background

Caching is currently not well-defined in logs.

## Aim

The aim is to improve the overall functionality of caching logs. Display which and how many files there are to be cached.

## Implementation

It is split between two different routes. The first is through the `grr_cache_repo` tool. When ran with verbose(`grr_cache_repo -v`) it displays which file is currently being cached and how many remain. Through running the server it displays only which file is being cached.

Closes #400.


environment.yml
```diff
@@ -51,3 +51,4 @@ dependencies:
   - dask=2023.3
   - toolz=0.12.0
   - markdown2=2.4.0
+  - tqdm=4.65.0
```
iordanivanov, line 54: Remove this as it is not used

iordanivanov (changes requested):

PR #420, opened 2023-03-30
Implement reference genome statistics

IvoTod:
## Background

Reference genomes should have statistics for nucleotide distribution and bi-nucleotide distribution

## Aim

Implement statistics for reference genomes.

## Implementation

Reference genomes have global and chromosome specific statistics. Chromosome specific statistics count nucleotides and global statistics count nucleotides and pairs. A new method has been added to the base statistic class `finish()`. It is an optional method which can be used for finalization of statistic data after data collection is complete. This is used to calculate the distribution in the new statistics.

### DIFF
--- a/dae/dae/genomic_resources/cli.py
+++ b/dae/dae/genomic_resources/cli.py
@@ -417,7 +417,7 @@ def _store_stats_hash(proto, resource):
     stats_dir = ResourceStatistics.get_statistics_folder()
     if stats_dir is None:
         logger.warning(
-            "Couldn't store stats hash for %s; unable to get stats dir", 
+            "Couldn't store stats hash for %s; unable to get stats dir",
             resource.resource_id)
         return False
     with proto.open_raw_file(
--- a/dae/dae/genomic_resources/tests/test_cli_stats.py
+++ b/dae/dae/genomic_resources/tests/test_cli_stats.py
@@ -490,7 +490,6 @@ def test_reference_genome_usage(tmp_path, mocker):
         new=ref_genome_length_mock
     )
     assert ref_genome_length_mock.call_count == 0
-    resource_path = os.path.join(str(tmp_path), "one")
     cli_manage([
         "resource-stats", "-r", "one", "-R", str(tmp_path), "-j", "1"
     ])


dae/dae/genomic_resources/reference_genome.py
```diff
@@ -25,6 +29,210 @@
...
+            )
+            self.nucleotide_counts = nucleotide_counts
+
+        self.nucleotide_distribution = {}
+        self.length = length
+        self.test = ""
+
+    def add_value(self, value):
+        self.length += 1
+        assert value in self.nucleotide_counts, value
```
lchorbadjiev, line 113: It happens that this assert fails on the real reference genomes. It is better to produce a warning message and skip the value:
```
        if value not in self.nucleotide_counts:
            logger.warning(
                "unexpected nucleotide <%s> in chromosome <%s>",
                value, self.statistic_id)
            return
```

dae/dae/genomic_resources/reference_genome.py
```diff
@@ -158,19 +366,20 @@ def get_all_chrom_lengths(self):
...
-    #         print(chrom, chrom_len)
-    #         logger.debug(
-    #             "Chromosome '%s' has length %s",
-    #             chrom, chrom_len)
-    #         i = 1
-    #         while i < chrom_len - region_size:
-    #             yield chrom, i, i + region_size - 1
-    #             i += region_size
-    #         yield chrom, i, None
+    def split_into_regions(self, region_size, chrom=None):
```
lchorbadjiev, line 369: I prefer this function to be external to the reference genome class.

dae/dae/genomic_resources/reference_genome.py
```diff
@@ -187,13 +396,39 @@ def get_sequence(self, chrom, start, stop):
...
+            length = self._index[chrom]["length"] - start + 1
+        else:
+            length = stop - start + 1
         line_feeds = 1 + length // self._index[chrom]["seqLineLength"]
 
         sequence = self._sequence.read(length + line_feeds).decode("ascii")
         sequence = sequence.replace("\n", "")[:length]
         return sequence.upper()
 
+    def pair_iter(self, region_size=1_000_000):
```
lchorbadjiev, line 409: I prefer that this function is external to reference genome class.

dae/dae/genomic_resources/reference_genome.py
```diff
@@ -268,10 +501,122 @@ def calc_info_hash(self):
         return "placeholder"
 
     def calc_statistics_hash(self) -> bytes:
```
lchorbadjiev, line 503: I would like to discuss if it is possible to make statistics calculations external to the resources. What I think is that we can create `ReferenceGenomeStatisticsBuilder`, which is initialized with the reference genome and organize statistics calculations.

The problem I would like to solve is that when we add statistics calculations in the resource class, the API becomes huge and irrelevant to the usage of the resource itself.

If we agree to work in this direction this task should be separated from the current PR.

dae/dae/genomic_resources/reference_genome.py
```diff
@@ -187,13 +396,39 @@ def get_sequence(self, chrom, start, stop):
...
+        Iterate and yield nucleotide pairs in the genome.
+
+        The return value is a triple of the chromosome, the left and the right
+        nucleotide.
+        """
+        self._sequence.seek(0)
+        last_chrom = None
+        last_nuc = None
+        for chrom, start, end in self.split_into_regions(region_size):
+            sequence = self.get_sequence(chrom, start, end)
```
lchorbadjiev, line 420: This reads the whole region into memory. We could probably add a new method, `fetch` to the reference genome class that returns an iterator over the specified sequence. And make the `get_sequence` method use this new `fetch` method.

dae/dae/genomic_resources/reference_genome.py
```diff
@@ -268,10 +501,122 @@ def calc_info_hash(self):
...
+            ) as outfile:
+                outfile.write(chrom_statistic.serialize())
+        return merged_statistics
+
+    @staticmethod
+    def _do_global_statistic(resource):
+        impl = build_reference_genome_from_resource(resource)
+        with impl.open():
+            statistic = GenomeStatistic(impl.chromosomes)
+            for pair in impl.pair_iter():
```
lchorbadjiev, line 608: This one task is going to iterate over the whole reference genome. This is going to be painfully slow.

Let us discuss this.

lchorbadjiev (changes requested):
I was about to approve this PR but found a stopper.

It seems to me that `ReferenceGenome._do_global_statistics` function is painfully slow. It tries to iterate over the whole reference genome, which is ~3GB. And it is done in a single task.

I am unsure why we could not use chromosome statistics to build the global statistics from them. I need to discuss this issue with you.

Another nuisance is the ~50 warnings increase the number of lint errors/warnings.

lchorbadjiev (changes requested):
I have found multiple issues with this PR.

I have tried to fix two of them in this commit:
https://github.com/iossifovlab/gpf/pull/420/commits/948050f9ee38f6c3c68b24f9c59fb562207f9347

We need some more tests that could catch these errors.

Also, there is one more issue I found. The reference genome statistics are recalculated each time I run `grr_manage repo-repair`.

lchorbadjiev (approved):

PR #436, opened 2023-04-20
Fix histogram throwing wrong error on wrong request #432

### LINKED ISSUE #432: [BUG] The histogram route returns 500 instead of 404
**What happened**: When the front-end makes requests it can often provide the wrong histogram id due to bugs. When this happens the result is error 500. 

**What you expected to happen**:  The usually expected error message is 404. Perhaps this would be more suitable

livovachkov:
## Background

The Histogram route is usually well protected and many checks are handled in order to be sure that the right data will be provided. However in some extreme cases in the front-end, there is a chance to be given the wrong measure. The exception is handled in a way that can be made more appropriate 

## Aim

Handle the request to return 400(Bad Request) instead of 500(Server Error) when the measure is not valid.

## Implementation

Closes #432.


wdae/wdae/measures_api/views.py
```diff
@@ -47,16 +47,19 @@ def post(self, request):
...
         dataset_id = data["datasetId"]
         dataset = self.gpf_instance.get_wdae_wrapper(dataset_id)
-
         assert dataset is not None
         assert dataset.phenotype_data is not None
-        assert "measure" in data
+        if "measures" in data \
+                and not dataset.phenotype_data.has_measure(data["measures"]):
+            return Response(status=status.HTTP_400_BAD_REQUEST)
+        assert "measures" in data
```
lchorbadjiev, line 55: Why do we change "measure" to "measures"? What is the `gpfjs` used in this request? How has it been working till now?

lchorbadjiev (changes requested):
We need to discuss the change from "measure" to "measures".

lchorbadjiev (approved):

PR #443, opened 2023-04-25
Grr browse improvements

### LINKED ISSUE #434: [ENH] Imporvements in `grr_browse`
- [x] dump GRR configuration in YAML format
- [x] make `--print-grr` default behavior

livovachkov:
## Background and aim

The improvements can be split into two subtasks. With migration into YAML, the idea is to transition the GRR definition into YAML format as the first task and the second is to make printing of the GRR default behavior.

## Implementation

Closes #434.

### DIFF
--- a/dae/dae/genomic_resources/cli.py
+++ b/dae/dae/genomic_resources/cli.py
@@ -1,4 +1,5 @@
 """Provides CLI for management of genomic resources repositories."""
+import json
 import os
 import sys
 import logging
@@ -794,7 +795,7 @@ def cli_browse(cli_args=None):
             print("Working with GRR definition:", definition_path)
         else:
             print("No GRR definition found, using the DEFAULT_DEFINITION")
-        pprint.pprint(definition)
+        print(yaml.dump(definition))
 
     repo = build_genomic_resource_repository(definition=definition)
     _run_list_command(repo, args)
--- a/dae/dae/genomic_resources/cli.py
+++ b/dae/dae/genomic_resources/cli.py
@@ -1,12 +1,10 @@
 """Provides CLI for management of genomic resources repositories."""
-import json
 import os
 import sys
 import logging
 import argparse
 import pathlib
 import copy
-import pprint
 
 from typing import Dict, Union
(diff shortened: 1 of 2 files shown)


lchorbadjiev (changes requested):
It is supposed that when we print the GRR definition, we should print it in YAML format, not as a dictionary.

Probably you could use something like `yaml.safe_dump`.

iordanivanov (changes requested):
~~Running `grr_browse` still prints the configuration as a dict instead of a YAML-formatted string.~~

Resolved.

dae/dae/genomic_resources/cli.py
```diff
@@ -789,12 +782,12 @@ def cli_browse(cli_args=None):
...
-            print("Working with GRR definition:", definition_path)
-        else:
-            print("No GRR definition found, using the DEFAULT_DEFINITION")
-        pprint.pprint(definition)
+
+    if definition_path is not None:
+        print("Working with GRR definition:", definition_path)
+    else:
+        print("No GRR definition found, using the DEFAULT_DEFINITION")
+    print(yaml.dump(definition))
```
lchorbadjiev, line 790: Always prefer `safe_dump` instead of `dump`.  Consider using `sort_keys=False` for keeping the GRR definition close to the original.

dae/dae/genomic_resources/tests/test_cli_browse.py
```diff
@@ -55,6 +55,10 @@ def test_cli_browse_with_grr_argument(repo_fixture, repo_def, capsys):
 
     assert err == ""
     assert out == \
+        "Working with GRR definition: " + str(repo_def) + "\n" \
```
lchorbadjiev, line 58: When constructing strings always prefer using f-strings. Don't use `+` for strings.

lchorbadjiev (changes requested):
Please check the changes requested in the comments.

lchorbadjiev (approved):

iordanivanov (approved):

PR #456, opened 2023-05-15
Region split utils

### LINKED ISSUE #419: Extract region splitting functionality into a separate util
Region splitting functionality can be found in multiple places:
- genomic_scores
- reference_genome
- annotate_vcf
- import tools

We should extract all of this functionality into a set of utilities that cover all our cases.

NOTES:
- Try extracting get_chromosome_length from GenomicPositionTable as a separate util for getting lengths of contigs not found in the reference genome, and then utilise it in our region splitting functions
- We expect to have two functions: one that creates a dictionary of (contig)->(length of contig) and another that converts such a dictionary to a set of regions, split by a given region size
- We should utilise the Region class in region_utils

### LINKED ISSUE #439: [BUG] reference genome new implementation of `get_sequence` occasionally produces bad results
Here we have a sequence of calls to the original and new implementations of `get_sequence`. In most cases, both implementation return equivalent results. But in the last case, the results are different by character.

```
OK    seq: chr4:1394548-1394549
OK    seq: chr4:1394550-1394550
OK    seq: chr4:1394548-1394549
OK    seq: chr4:1394550-1394559
OK    seq: chr4:1394557-1394559
OK    seq: chr4:1394557-1394557
OK    seq: chr4:1394554-1394555
OK    seq: chr4:1394556-1394712
OK    seq: chr4:1394557-1394557
====================================================================================================
ERROR seq: chr4:1394558-1394835
original:  ACACGTGCCCATGTGGAGTGCCCGCCTGCTCATGTGCCCATGTGGAGTGCCCGCCTGCTCACACATGTCGATGCGGAGTGCCCGCCTGCTCACACATGCCCATGTGGAGTGCCCGCCTGCTCACACGTGCCCATGTGGAGTGCCCGCCTGCTCACACACGTGTCCATGTGGAGTGCCCACCTGCTCATGTGCCCATGTGGAGTGCCCACCTGCTCACATGTGCCGATGTGGAGTGCCACCTGCTCACACACGTGCCCATGTGGAGTGCCCGCCTGCTC
new impl: ACACGTGCCCATGTGGAGTGCCCGCCTGCTCATGTGCCCATGTGGAGTGCCCGCCTGCTCACACATGTCGATGCGGAGTGCCCGCCTGCTCACACATGCCCATGTGGAGTGCCCGCCTGCTCACACGTGCCCATGTGGAGTGCCCGCCTGCTCACACACGTGTCCATGTGGAGTGCCCACCTGCTCATGTGCCCATGTGGAGTGCCCACCTGCTCACATGTGCCGATGTGGAGTGCCACCTGCTCACACACGTGCCCATGTGGAGTGCCCGCCTGCT
====================================================================================================
```

IvoTod:
## Background

Regions were handled individually in different modules, resulting in duplicate code.

## Aim

Add new generic utilities for chromosome splitting and acquiring length to be used in said modules.

## Implementation
Newly added utilities are `get_chromosome_length` and `split_into_regions`. They are used in genomic scores and genomic position table. This PR also includes a fix for the bugged get_sequence method

## Related issues

Closes #419.
Closes #439.

### DIFF
--- a/dae/dae/utils/regions.py
+++ b/dae/dae/utils/regions.py
@@ -34,10 +34,10 @@ def split_into_regions(
     regions = []
 
     current_start = 1
-    while current_start < chrom_length:
-        end = min(chrom_length, current_start + region_size)
+    while current_start < chrom_length + 1:
+        end = min(chrom_length, current_start + region_size - 1)
         regions.append(Region(chrom, current_start, end))
-        current_start = end
+        current_start = current_start + region_size
 
     return regions
 
--- a/dae/dae/utils/regions.py
+++ b/dae/dae/utils/regions.py
@@ -36,7 +36,7 @@ def split_into_regions(
     current_start = 1
     while current_start < chrom_length:
         end = min(chrom_length, current_start + region_size)
-        regions.append(Region(chrom, current_start))
+        regions.append(Region(chrom, current_start, end))
         current_start = end
 
     return regions
--- a/dae/dae/utils/regions.py
+++ b/dae/dae/utils/regions.py
@@ -2,6 +2,8 @@ from __future__ import annotations
 from collections import defaultdict
 import copy
 
+import pysam
+
 
 def bedfile2regions(bed_filename):
     """Transform BED file into list of regions."""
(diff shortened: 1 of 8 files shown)


dae/dae/utils/regions.py
```diff
@@ -25,6 +27,62 @@ def regions2bedfile(regions, bed_filename):
...
+    current_start = 1
+    while current_start < chrom_length + 1:
+        end = min(chrom_length, current_start + region_size - 1)
+        regions.append(Region(chrom, current_start, end))
+        current_start = current_start + region_size
+
+    return regions
+
+
+def get_chromosome_length(
```
lchorbadjiev, line 45: I would like to have `tabix` somewhere in the name of this function. Probaly `get_tabix_file_chromosome_length`, or `get_tabix_file_chrom_len` or something similar.

lchorbadjiev (approved):
Please consider renaming `get_chromosome_lenght` util function before merging.

PR #473, opened 2023-06-02
Fix chrom mapping through genomic scores.

IvoTod:
## Background
The new utility for getting chromosome lengths interfaces directly with pysam files. The tables in the genomic scores can be configured to map chromosomes internally. The score is dependent on the table to receive an accurate list of chromosomes, but it receives a transformed list. This creates an error when using chrom mapping because the utility is unaware of the mapped chromosomes

## Aim
We should be able to use the utility with a chrom mapping set.

## Implementation
I have implemented new methods to the table and score for mapping a specific chromosome and a new getter for the original, unmodified list of chromosomes. This is a bit of a band aid solution to the problem and further improvements are open to discussion.

### DIFF
--- a/dae/dae/genomic_resources/genomic_scores.py
+++ b/dae/dae/genomic_resources/genomic_scores.py
@@ -403,7 +403,7 @@ class GenomicScoreImplementation(
         regions = []
         ref_genome_id = self.resource.get_labels().get("reference_genome")
         ref_genome = self._get_reference_genome_cached(grr, ref_genome_id)
-        for chrom in self.score.get_all_chromosomes():
+        for chrom in self.score.get_all_chromosomes_unmapped():
             if ref_genome is not None:
                 chrom_length = ref_genome.get_chrom_length(chrom)
             else:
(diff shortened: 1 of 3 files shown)


iordanivanov (approved):

iordanivanov:
As an idea for future improvement - perhaps chrom_order and chrom_order_unmapped could be derived from the chrom_map and rev_chrom_map dictionaries, in order to avoid a potential disconnect between the two chrom_order sets

dae/dae/genomic_resources/genomic_scores.py
```diff
@@ -1030,6 +1035,12 @@ def get_all_chromosomes(self):
 
         return self.table.get_chromosomes()
 
+    def map_chromosome(self, chromosome):
+        return self.table.map_chromosome(chromosome)
+
+    def unmap_chromosome(self, chromosome):
+        return self.table.unmap_chromosome(chromosome)
+
```
lchorbadjiev, line 1043: I prefer to remove these methods and use the table methods.

lchorbadjiev (approved):
Please remove the `map_chromosome` and `unmap_chromosome` methods from the genomic score and use the corresponding methods from the table.

PR #487, opened 2023-06-30
Federation token api update

### LINKED ISSUE #480: [ENH] Update federation token api
Remove client secret and id from request body and allow updating the token's name.

Remove client id and secret from api GET request.
Implement PUT request that updates a given token's name.

IvoStefanov:
## Background

update federation credentials api to support updating existing credentials and remove some of the sent information

## Aim

add put request for singular credential name update
remove credential secret and id from get request body
add unit tests for the whole api

Closes #480

### DIFF
--- a/wdae/wdae/users_api/views.py
+++ b/wdae/wdae/users_api/views.py
@@ -502,8 +502,6 @@ class FederationCredentials(views.APIView):
         for app in apps:
             res.append({
                 "name": app.name,
-                "client_id": app.client_id,
-                "client_secret": app.client_secret,
             })
         return Response(res, status=status.HTTP_200_OK)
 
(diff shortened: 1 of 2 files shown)


wdae/wdae/users_api/views.py
```diff
@@ -543,10 +541,43 @@ def delete(self, request):
...
+            return Response(status=status.HTTP_401_UNAUTHORIZED)
+        if not get_application_model() \
+                .objects \
+                .filter(name=request.data.get("name")) \
+                .exists():
+            return Response(status=status.HTTP_400_BAD_REQUEST)
+        if (get_application_model()
+                .objects
+                .filter(name=request.data.get("new_name"))
+                .exists()):
```
iordanivanov, line 571: Prefer backslash for newline over parentheses

wdae/wdae/users_api/views.py
```diff
@@ -543,10 +541,43 @@ def delete(self, request):
...
+            return Response(status=status.HTTP_400_BAD_REQUEST)
+        app = get_application_model().objects.get(
+            name=request.data.get("name")
+        )
+        if not user.id == app.user_id:
+            return Response(status=status.HTTP_401_UNAUTHORIZED)
+        app.name = request.data.get("new_name")
```
iordanivanov, line 578: You should add a check that `new_name` is both present in the request data and has a value (is not None). If this check fails, return a 400 response.

iordanivanov: After that you could convert this `.get` access into bracket notation (i.e. `request.data["new_name"]`) - this clearly conveys that we expect `new_name` to be present in the dict since this will throw a `KeyError` if it is not found (in contrast to `.get`, which will just return a default value of `None`)

iordanivanov (changes requested):

iordanivanov (approved):

PR #490, opened 2023-07-04
Support for case insensitive gene symbols for the Gene models and AGP…

### LINKED ISSUE #486: [BUG] Support case insensitive gene symbols in the GPF
**What happened**: When querying gene symbols from the gene models and the autism gene profiles database, among other places, the results are case-sensitive. For example, querying “CHD8” returns a different result than querying “chd8” or “Chd8” in the gene models.

**What you expected to happen**: The gene symbols should be case insensitive so that they return the same result regardless of the case variation. For example, querying “CHD8”, “chd8”, or “Chd8” should all return the same gene model and autism gene profile. The is the backend to be case insensitive but also preserve the case sensitivity according to its database model.

livovachkov:
… DB #486

## Background

The backend currently is case sensitive for gene symbols in a lot of places such as the gene models and the AGP.

## Aim

Make it as case-insensitive as possible, there is a possible issue that may be related to the impala(?).

## Implementation

Closes #486.

### DIFF
--- a/wdae/wdae/genomes_api/tests/test_gene_models_api.py
+++ b/wdae/wdae/genomes_api/tests/test_gene_models_api.py
@@ -40,11 +40,14 @@ def test_get_nonexistant_gene_transcripts(anonymous_client):
     assert response
     assert response.status_code == 404
 
+
 def test_get_case_insensitive_gene(anonymous_client):
     response = anonymous_client.get("/api/v3/genome/gene_models/default/cHd8")
 
     assert response
+    assert response.data["gene"] == "CHD8"
     assert response.status_code == 200
+
 
 def test_search_gene_symbols(anonymous_client):
     response = anonymous_client.get("/api/v3/genome/gene_models/search/CHD8")
--- a/wdae/wdae/genomes_api/tests/test_gene_models_api.py
+++ b/wdae/wdae/genomes_api/tests/test_gene_models_api.py
@@ -40,6 +40,11 @@ def test_get_nonexistant_gene_transcripts(anonymous_client):
     assert response
     assert response.status_code == 404
 
+def test_get_case_insensitive_gene(anonymous_client):
+    response = anonymous_client.get("/api/v3/genome/gene_models/default/cHd8")
+
+    assert response
+    assert response.status_code == 200
 
 def test_search_gene_symbols(anonymous_client):
     response = anonymous_client.get("/api/v3/genome/gene_models/search/CHD8")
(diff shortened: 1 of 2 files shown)


dae/dae/autism_gene_profile/db.py
```diff
@@ -80,7 +80,9 @@ def get_agp(self, gene_symbol) -> Optional[AGPStatistic]:
         """
         table = self.agp_table
         query = table.select()
-        query = query.where(table.c.symbol_name == gene_symbol)
+        query = query.where(
+            func.lower(table.c.symbol_name) == func.lower(gene_symbol)
```
iordanivanov, line 84: You could use the native Python `.lower()` for `gene_symbol` since it's a simple `str` instance

dae/dae/genomic_resources/gene_models.py
```diff
@@ -491,7 +492,11 @@ def gene_names(self):
         return list(self.gene_models.keys())
 
     def gene_models_by_gene_name(self, name):
-        return self.gene_models.get(name, None)
+        pattern = re.compile(re.escape(name) + "$", re.IGNORECASE)
+        for key in self.gene_models:
+            if pattern.match(key):
+                return self.gene_models[key]
+        return None
```
iordanivanov, line 499: I'm not really sure about this code... it turns a O(1) function into an O(n) one.

wdae/wdae/genomes_api/views.py
```diff
@@ -16,23 +17,23 @@ def get(self, request):
...
+                    "transcripts": [],
+                }
+                for tr in transcripts:
+                    response_data["transcripts"].append(
+                        self.transcript_to_dict(tr)
+                    )
+
+                return Response(
+                    response_data,
+                    status=status.HTTP_200_OK,
```
iordanivanov, line 34: Similar concern here.

iordanivanov (commented):
I'm not very sure about this approach. Let's discuss this further.

wdae/wdae/genomes_api/tests/test_gene_models_api.py
```diff
@@ -40,6 +40,11 @@ def test_get_nonexistant_gene_transcripts(anonymous_client):
     assert response
     assert response.status_code == 404
 
+def test_get_case_insensitive_gene(anonymous_client):
+    response = anonymous_client.get("/api/v3/genome/gene_models/default/cHd8")
+
+    assert response
+    assert response.status_code == 200
```
iordanivanov, line 47: Validate the response data is actually the CHD8 gene, not just the 200 status code

iordanivanov (changes requested):

iordanivanov (approved):

PR #519, opened 2023-08-22
Fix for gene sets downloads in remote download

livovachkov:
## Background

Previously, when retrieving gene sets from a remote source, the "queryData" parameter was being set incorrectly on the root object passed to the expand_gene_set utility. This was causing issues when trying to download gene sets from remote sources.

## Aim

This PR fixes the bug where queryData was being set incorrectly. Now, expand_gene_set will receive the proper "queryData" when processing gene sets from remote sources.

## Implementation

Identified where queryData was being set on the root object instead of the appropriate child object
Updated the code to set queryData on the correct child object before passing data to expand_gene_set

### DIFF
--- a/wdae/wdae/utils/expand_gene_set.py
+++ b/wdae/wdae/utils/expand_gene_set.py
@@ -1,13 +1,11 @@
-from typing import Any
 from gpf_instance.gpf_instance import get_wgpf_instance
 
 from datasets_api.permissions import IsDatasetAllowed
 
 
-def expand_gene_set(data: Any, user: dict) -> dict:
+def expand_gene_set(data: dict, user: dict) -> dict:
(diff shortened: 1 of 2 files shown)


wdae/wdae/utils/expand_gene_set.py
```diff
@@ -28,8 +28,10 @@ def expand_gene_set(data: dict, user: dict) -> dict:
     return data
 
 
-def expand_gene_syms(data: dict, user: dict) -> Any:
+def expand_gene_syms(data: Any, user: dict) -> dict:
     """Expand gene set symbols."""
+    data = dict(data)
```
IvoTod, line 33: is this necessary?

IvoTod (approved):

IvoTod (approved):

PR #523, opened 2023-08-28
VCF reannotation

IvoTod:
## Background

Reannotation support was necessary for the `annotate_vcf.py` tool after it was done in the `annotate_columns.py` tool.

## Aim
Add support for `--reannotation` flag in `annotate_vcf.py`.

## Implementation

Reannotation in `annotate_vcf.py` was made similarly to the way it works in `annotate_columns.py`. This PR also fixes a bug with `annotate_columns.py`, where if an annotator with internal attributes in the old pipeline is deleted in the new pipeline, it will attempt to access them where it cannot and crash. Also included is a change to the default attributes of `EffectAnnotator`, which used to be internal by default, but were changed at some point (most likely by accident).

### DIFF
--- a/dae/dae/annotation/annotate_columns.py
+++ b/dae/dae/annotation/annotate_columns.py
@@ -198,7 +198,8 @@ def annotate(
                         record_to_annotatable.build(record)
                     )
 
-                record.update(annotation)
+                for col in annotation_columns:
+                    record[col] = annotation[col]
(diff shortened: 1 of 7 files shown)


lchorbadjiev (approved):

iordanivanov (approved):
I'd like to see a unit test for the deleted internal attributes bugfix, but no complaints otherwise

PR #529, opened 2023-09-01
Add default permission denied prompt

livovachkov:
## Background

Currently, if the permission denied prompt config file is missing - it displays a blank page, without any warning or error.

## Aim

The aim is to provide a default permission denied config if one is missing.


iordanivanov (approved):

iordanivanov:
I'd like to see `default_config` renamed to `default_prompt`.

PR #534, opened 2023-09-13
Remote study wrapper now doesn't round the response download

livovachkov:
## Background

Currently, when the Genotype browser download is made, the scores it downloads are not properly truncated. This makes the behavior inconsistent with the non-remote instance which rounds the float values up to the third number.

## Aim

To provide more consistent behavior.

## Implementation

Note: Tests have to be split on another issue due to the fixtures missing scores.

### DIFF
--- a/wdae/wdae/studies/study_wrapper.py
+++ b/wdae/wdae/studies/study_wrapper.py
@@ -589,6 +589,10 @@ class RemoteStudyWrapper(StudyWrapperBase):
             kwargs, reduce_alleles=False
         )
 
+        for source in sources:
+            if "format" in source:
+                del source["format"]
+
         def get_source(col: dict[str, Any]) -> Any:
             res = col["source"]
             if "role" in col:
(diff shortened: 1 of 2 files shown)


IvoTod (changes requested):
The remote query response could have been deleting it's columns' formats because it relied on them to come pre-formatted from the remote. We should check if this is still the case or if it is broken and fixable. I'm going to request changes for this PR and think we should investigate this a little further before continuing with the current changes.

wdae/wdae/studies/study_wrapper.py
```diff
@@ -592,7 +589,11 @@ def query_variants_wdae_streaming(
             kwargs, reduce_alleles=False
         )
 
-        def get_source(col: dict[str, Any]) -> Any:
+        for source in sources:
+            if "format" in source:
+                del source["format"]
+
+        def get_source(col):
```
IvoTod, line 596: Restore the typing on this function

IvoTod (changes requested):

IvoTod (approved):

PR #538, opened 2023-09-20
Fix family combining

IvoTod:
## Background

There was an issue with family type tags in studies that are in groups.

## Aim

Find and fix the problem with the tags.

## Implementation

The problem was that when we initialize a genotype data group, we create its `FamiliesData` on the spot, by using `.combine` with all of its child genotype datas' `FamiliesData`.  `.combine` collects all of the shared family IDs between two instances of `FamiliesData` and merges each. Before doing that, all of the families from the first get pasted into a dictionary, and then all of the families from the second get pasted over. This exact step is erroneous, because any family that isn't shared between the two instances is just pasted over as a reference, whereas the shared families are merged into new instances. 
When we tag a family that has not been merged, its tags will persist and the tagging will be broken and leak.

### DIFF
--- a/dae/dae/pedigrees/tests/test_combine_families.py
+++ b/dae/dae/pedigrees/tests/test_combine_families.py
@@ -9,6 +9,7 @@ from dae.pedigrees.family_tag_builder import FamilyTagsBuilder
 
 from dae.testing import setup_pedigree
 
+
 @pytest.fixture
 def ped_a(tmp_path: pathlib.Path) -> FamiliesData:
     ped_path = setup_pedigree(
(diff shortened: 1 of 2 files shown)


dae/dae/pedigrees/tests/test_combine_families.py
```diff
@@ -184,3 +216,46 @@ def test_combine_families_generated_and_missing(
...
+    assert len(ped_f["f2"].tags.intersection(["tag_family_type"])) == 0
+    assert len(ped_g["f1"].tags.intersection(["tag_family_type"])) == 0
+    print(ped_a["f1"].persons["f1.dad"]._attributes)
+    print(ped_f["f2"].persons["f2.dad"]._attributes)
+    print(ped_g["f1"].persons["f1.dad"]._attributes)
+    print(ped_g["f3"].persons["f3.dad"]._attributes)
+    print(ped_a["f1"].tags)
+    print(ped_f["f2"].tags)
+    print(ped_g["f1"].tags)
+    print(ped_g["f3"].tags)
```
lchorbadjiev, line 257: Remove debug prints.

lchorbadjiev (approved):

PR #550, opened 2023-10-11
Remote genomic scores

IvoTod:
## Background
Since we switched to genomic scores being tied to annotation, setting up local genomic scores to substitute this as a feature has become difficult.

## Aim
Implement remote genomic scores.

## Implementation
Genomic scores are represented by a `ScoreDesc` class, so these are now exposed on a new view on the API. The remote genomic scores DB fetches all of the `ScoreDesc` classes on the remote and recreates them locally, prefixing an ID to the description. Because this included changes in the WDAE GPF instance, this made a lot of lint errors resurface, so this is also addressed in this PR.

### DIFF
--- a/dae/dae/gene/scores.py
+++ b/dae/dae/gene/scores.py
@@ -20,7 +20,7 @@ class ScoreDesc:
     resource_id: str
     score_id: str
     source: str
-    destination: str
+    name: str
     hist: Optional[Union[NumberHistogram, CategoricalHistogram, NullHistogram]]
     description: str
     help: str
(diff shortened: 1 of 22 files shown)


dae/dae/gene/scores.py
```diff
@@ -19,11 +20,50 @@ class ScoreDesc:
     resource_id: str
     score_id: str
     source: str
-    destination: str
-    hist: Union[NumberHistogram, CategoricalHistogram, NullHistogram]
+    name: str
+    hist: Optional[Union[NumberHistogram, CategoricalHistogram, NullHistogram]]
```
lchorbadjiev, line 24: Why is histogram optional?

PR #569, opened 2023-11-13
Agp config full test

IvoStefanov:
## Background
Test agp generation with bad order and default dataset fields.

## Aim
Create a more comprehensive test that uses most of the agp generation pipeline and not mocks.

## Implementation
Use t4c8 instance to test agp generation properly.
Test generated agp genes data.
Test if generated agp with incomplete config throws error.

## Related issues
Fix using invalid order in agp config. Create a test for this.
Fix ewa files starting with empty line because they are not trimmed. All agp tests test this


lchorbadjiev (approved):
Consider reorganizing AGP config fixtures before merging.

Issue #564, opened 2023-11-09
[ENH] Reimplement remote phenotype browser downloads


IvoTod:
**Is your feature request related to a problem? Please describe.**
The backend code for phenotype browser downloads was rewritten, but the remote download was not taken into account initially. Currently there is a workaround applied for remote downloads.
Before, the download worked with a method in the pheno DB called `get_values_streaming_csv` which returned a generator with string CSV lines. This method was changed to `get_people_measure_values` and now returns a list of dictionaries for every row. This list gets transformed into a CSV in the view itself. Because the remote REST API client relies on the view, it currently cannot return the same dictionary in the remote `get_people_measure_values` method.

**Describe the solution you'd like**
We should implement a new view in the API, which should stream the rows from `get_people_measures_values` as JSONs and use it in the remote phenotype DB through the REST API client.

PR #575, opened 2023-11-28
Remote pheno measure values api

IvoStefanov:
## Background
The backend code for phenotype browser downloads was rewritten, but the remote download was not taken into account initially. Currently there is a workaround applied for remote downloads.
Before, the download worked with a method in the pheno DB called get_values_streaming_csv which returned a generator with string CSV lines. This method was changed to get_people_measure_values and now returns a list of dictionaries for every row. This list gets transformed into a CSV in the view itself. Because the remote REST API client relies on the view, it currently cannot return the same dictionary in the remote get_people_measure_values method.

## Aim
We should implement a new view in the API, which should stream the rows from get_people_measures_values as JSONs and use it in the remote phenotype DB through the REST API client.

## Related issues
Background and aim provided by the issue: #564

### DIFF
--- a/wdae/wdae/pheno_browser_api/views.py
+++ b/wdae/wdae/pheno_browser_api/views.py
@@ -139,7 +139,6 @@ class PhenoMeasuresDownload(QueryDatasetView):
     def csv_value_iterator(
         self,
         dataset: Union[RemoteStudyWrapper, StudyWrapper],
-        columns: dict[str, str],
         measure_ids: list[str]
     ) -> Generator[str, None, None]:
         """Create CSV content for people measures data."""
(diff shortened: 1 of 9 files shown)


wdae/wdae/pheno_browser_api/views.py
```diff
@@ -191,19 +191,54 @@ def post(self, request: Request) -> Response:
...
 
         response["Content-Disposition"] = "attachment; filename=measures.csv"
         response["Expires"] = "0"
         return response
+
+
+class PhenoMeasureValues(QueryDatasetView):
+
+    def post(self, request: Request) -> Response:
+        """Return a CSV file stream for measures."""
```
IvoTod, line 207: Adjust this documentation

IvoTod (approved):

PR #577, opened 2023-12-05
Instrument value tables

IvoTod:
## Background
Measure downloads were still slow after memory optimization, so we decided to use a cache table approach.

## Aim
Implement a cache table for measure downloads

## Implementation
Due to SQL limitations on size and amount of joins, new DB tables have been added, `instruments` and `measures`. The new `measures` table is meant to replace the old one in time since it is mostly the same and contains an instrument FK instead of a name and a `db_column_name`. Measure IDs as columns are now converted to a more DB-friendly format and are stored in the `db_column_name`. Same goes for instrument names, which are stored in `instruments.table_name`.
The other new tables are instrument measure value tables, one for each instrument in the DB. These have person ID, family ID, role and respective measure values as columns. Each of these contains part of the download. The final download is made by doing a `FULL OUTER JOIN` on all of these instrument value tables, so that no values are omitted.
A new tool has been added - `build_pheno_measures` which prepares all these new tables.


lchorbadjiev (approved):
Consider removing the usage of `print` before merging.

Move the `build_pheno_measures` script to the `dae.pheno` package and define it in `[entry_points]` section instead of `[scripts]` section of `setup.py`.

PR #587, opened 2024-01-05
Remote pheno images

IvoTod:
## Background
Phenotype browser images had an implementation, where the links were matched to lead directly to the remote server's static files route, but this ended up not working because of cross site policies. The images for remotes needed a different implementation to work.

## Aim
Write an alternative solution to displaying remote phenotype browser images.

## Implementation
The policy is circumvented by fetching the images in the backend via the remote REST API client. A new view has been made to serve raw images from remotes. The old implementation of matching the base URL to the remote has been removed and now the base URL leads to the correct remote images view with the correct remote ID. Remote clients were also loaded in a list in the WDAE GPF instance and later mapped to the loaded in remote studies. This worked until now because almost every site functionality is dependent on a dataset ID. Because images are tied to a specific remote and specific pheno directory and not a study, this old way of getting REST clients suddenly stopped making sense. REST clients are now loaded in a dictionary, mapped to their remote IDs.

### DIFF
--- a/wdae/wdae/remote/remote_phenotype_data.py
+++ b/wdae/wdae/remote/remote_phenotype_data.py
@@ -242,12 +242,18 @@ class RemotePhenotypeData(PhenotypeData):
     def get_regressions(self):
         return self.rest_client.get_regressions(self.remote_dataset_id)
 
+    @staticmethod
+    def _extract_pheno_dir(url: str) -> str:
+        """Extract the pheno directory from a measures info URL."""
+        url = url.strip("/")
+        pheno_folder = url[url.rindex("/") + 1:]
+        return pheno_folder
+
     def get_measures_info(self):
         output = self.rest_client.get_browser_measures_info(
             self.remote_dataset_id
         )
-        remote_base = output["base_image_url"][:-1]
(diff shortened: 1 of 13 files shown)


wdae/wdae/remote/remote_phenotype_data.py
```diff
@@ -241,8 +246,12 @@ def get_measures_info(self):
...
             self.remote_dataset_id
         )
-        output["base_image_url"] = \
-            f"/{self.rest_client.gpf_prefix}/{output['base_image_url']}"
+        remote_base = output["base_image_url"][:-1]
+        pheno_folder = remote_base[remote_base.rindex("/") + 1:]
+        output["base_image_url"] = (
+            "/api/v3/pheno_browser/remote_images/"
+            f"{self.rest_client.remote_id}/{pheno_folder}/"
+        )
```
lchorbadjiev, line 254: The parsing of `base_image_url` seems frigile. 

What happens when `base_image_url` ends with `//` or if it has no `/` at the end.

Could you please create a function that does this transformation and add tests for it.

lchorbadjiev (approved):
Consider making the transformation of `base_image_url` more robust before merging.

PR #589, opened 2024-01-08
Add new tag filter functionality

IvoTod:
## Background
Pedigree tag functionality was recently expanded in the frontend to support and/or and tag exclusion. The download however needs to be updated.

## Aim
Update the backend download handling of pedigrees to support the new tag filtering.

## Implementation
The old download was a GET request, due to the more complicated data being sent, this is now a POST request. The requests expects either an empty body to download all families or the following format.
```json
{
    "tagsQuery": {
        "orMode": true/false,
        "includeTags": [],
        "excludeTags": []
    }
}
```
A new utility was written to handle this tag query format. This format could be made more robust for future use in different parts of the system.

### DIFF
--- a/wdae/wdae/common_reports_api/views.py
+++ b/wdae/wdae/common_reports_api/views.py
@@ -1,20 +1,22 @@
+from typing import Optional, Any
 from django.http.response import StreamingHttpResponse
 from rest_framework import status
+from rest_framework.request import Request
 from rest_framework.response import Response
 
 from query_base.query_base import QueryDatasetView
 
 from utils.query_params import parse_query_params
-from dae.pedigrees.family import FamilyTag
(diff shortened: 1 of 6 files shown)


wdae/wdae/common_reports_api/views.py
```diff
@@ -129,9 +131,70 @@ def post(self, request):
...
+                    return True
+            return False
+
+        for tag in include_tags:
+            if not check_tag(family, tag):
+                return False
+        for tag in exclude_tags:
+            if check_tag(family, tag):
+                return False
+        return True
```
lchorbadjiev, line 157: Probably, this method should be removed.

lchorbadjiev (approved):

PR #598, opened 2024-01-30
Pheno registry

IvoTod:
## Background
As an initial step towards improving our pheno DB loading time, we decided to refactor the entire pheno codebase.

## Aim
Refactor and clean up the pheno modules.

## Implementation
The old `PhenoDb` class is now replaced by `PhenoRegistry`. The old `DbManager` class is now renamed to `PhenoDb`, and the modules have been changed accordingly. `pheno/pheno_db.py` is now `pheno/pheno_data.py`. The new `PhenoRegistry` class supports locks for avoiding multiple loads of the same phenotype study and is more decoupled from the GPF instance. The entire pheno codebase now uses the new `get_people_measure_values` API, instead of the old one, where there was a different method for different scenarios, usually with a dataframe and non dataframe version.


dae/dae/pheno/tests/test_default_filters.py
```diff
@@ -1,8 +1,9 @@
 # pylint: disable=W0621,C0114,C0116,W0212,W0613,too-many-lines
 import pytest
-from dae.pheno.pheno_db import PhenotypeStudy
+from dae.pheno.pheno_data import PhenotypeStudy
 
 
+@pytest.mark.skip(reason="Filters unsupported")
```
lchorbadjiev, line 6: Probably, we should remove default filtering.

lchorbadjiev (approved):

PR #603, opened 2024-02-20
Fix GP counting bug

IvoTod:
## Background
A new GP counting bug was found with the SHANK2 gene on the e2e build.

## Aim
Investigate and fix the counting bug.

## Implementation
The logic for whether a variant contributes to the count would take the allele's `effect_genes` and make a map of symbol to effect. In our example with the bug, the variant's effect gene list would look like this: `[SHANK2:frame-shift, SHANK2:non-coding-intron]`. In our example the statistic should count, because it is a frame shift, but after making the map, `frame-shift` gets overwritten by `non-coding-intron`, thus skipping the count. The fix made changes the dictionary map from symbol to effect -> symbol to set of effects.

### DIFF
--- a/dae/dae/gene_profile/generate_gene_profile.py
+++ b/dae/dae/gene_profile/generate_gene_profile.py
@@ -74,11 +74,16 @@ def add_variant_count(
         if effect_types is not None:
             skip = True
             for allele in variant.alt_alleles:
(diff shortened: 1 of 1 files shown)


lchorbadjiev (commented):
LGTM

lchorbadjiev (approved):
LGTM

PR #610, opened 2024-03-29
update response transformer genes to be only ones with worst effect

IvoStefanov:
## Background
Genes shown in genotype browser effect column are not only the ones that have the worst effect shown in the same column.

## Aim
Update response transformer genes to be the only ones that have the worst effect.

### DIFF
--- a/dae/dae/effect_annotation/effect.py
+++ b/dae/dae/effect_annotation/effect.py
@@ -847,6 +847,21 @@ def gene_effect_get_worst_effect(
     return ",".join([gene_effects.worst])
 
 
+def gene_effect_get_genes_worst(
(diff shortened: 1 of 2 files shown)


dae/dae/effect_annotation/effect.py
```diff
@@ -847,6 +847,21 @@ def gene_effect_get_worst_effect(
     return ",".join([gene_effects.worst])
 
 
+def gene_effect_get_genes_worst(
+    gene_effects: Optional[AlleleEffects]
+) -> str:
+    """Return command separted list of genes."""
```
iordanivanov, line 853: Error in comment, perhaps you meant "semicolon-separated"?

iordanivanov (approved):

PR #623, opened 2024-04-30
Pedigree only import support

IvoTod:
## Background
When using import projects, importing studies with only pedigree data was erroneous. We want to support this to import studies from phenotype data for previewing pheno data only.

## Aim
Fix and add tests for pedigree only imports

## Implementation
Duckdb imports were missing a table in one case and always assumed variants tables would be present, despite actually being programmed to handle cases without variants tables.


lchorbadjiev (approved):
mop up pylint and ruff errors before merging

PR #628, opened 2024-05-16
Annotate doc UI

IvoStefanov:
## Background
Annotate doc ui is less than ideal.

## Aim
Update ui.

## Implementation
Rewrite annotate doc jinja template.


iordanivanov (changes requested):
Remove the two .html files

iordanivanov (approved):

PR #642, opened 2024-05-29
Add pheno import parallelization

IvoTod:
## Background
When updating the pheno browser import scripts, the import was noticeably slow when performing the browser regression and image generation steps. We decided to attempt to parallelize the process using the `TaskGraph`.

## Implementation
For every single measure in every single instrument, a single task is created for calculating and generating regressions and images. The code had to be moved around a bit, due to connections not being serializable and a workaround has been implemented for workers to be able to access the DB due to DuckDB's read/write limitations: when creating workers, a temporary copy for reading is created. A new utility has been added to the task graph toolkit for running a task graph and getting direct results from the tasks.


lchorbadjiev (approved):
Make sure to mop up pylint, mypy, and ruff errors before merging.

PR #656, opened 2024-06-17
External demo annotator

IvoTod:
## Background
We decided to create a demo annotator to begin experimenting with integrating third party tools into our annotation pipeline.

## Aim
Create a demo annotator with an external tool which measures the length of a given annotatable.

## Implementation
The annotator classes used for this are called adapters and are built around calling another tool via `subprocess`. The first adapter uses a custom demo tool made for calculating the length of an annotatable.
This is in a new package in GPF called `external_demo_annotator` and the tool is available in the CLI when installed as `annotate_length` and the annotators can be added to a pipeline. The 2 annotator types added are `external_demo_annotator` and `external_demo_stream_annotator`


lchorbadjiev (approved):

lchorbadjiev:
Please consider squashing and reorganizing commits before merging

PR #674, opened 2024-07-08
VEP annotators

IvoTod:
## Background
After experimenting with external annotators we started working on adding real tools to our annotation pipeline.

## Aim
Add support for annotation with Ensembl's VEP tool.

## Implementation
Two VEP annotators have been created, one relies on a VEP cache being prepared, the other uses our own genome and gene model genomic resources. Along with this, the older external annotators have been refactored and a few lint rules which overlap with RUFF have been disabled.


external_demo_annotator/demo_annotator/vep_annotator.py
```diff
@@ -0,0 +1,398 @@
...
+            subprocess.run(args, check=True)
+            out_file.flush()
+            self.read_output(out_file, contexts)
+
+        self.aggregate_attributes(contexts)
+
+        return contexts
+
+
+class DemoVEPGTFAnnotator(VEPAnnotatorBase):
```
lchorbadjiev, line 302: We should remove the `Demo` from the name of this annotator. Please rename to `VEPGtfAnnotator` or `VEPFastaGtfAnnotator`?

external_demo_annotator/demo_annotator/vep_annotator.py
```diff
@@ -0,0 +1,398 @@
...
+
+            context["worst_consequence"] = [sorted(
+                the_consequences, key=lambda x: CONSEQUENCES[x],
+            )[0]]
+            context["highest_impact"] = [sorted(
+                context["impact"], key=lambda x: IMPACTS[x],
+            )[0]]
+
+
+class DemoVEPCacheAnnotator(VEPAnnotatorBase):
```
lchorbadjiev, line 258: We should remove `Demo` from the name of this annotator. Please rename to `VEPCacheAnnotator`

lchorbadjiev (approved):
We should rename the packages for this plugin. 

The base directory should be `external_vep_annotator`.

Inside we could have `vep_annotator` and `demo_annotator` packages.

PR #680, opened 2024-07-12
Improvement and fixes in serialisation of gene models in GTF format

lchorbadjiev:
## Background

When serializing gene models into GTF format, we need to produce a `phase` field for
all coding sequence features - CDS, start_codon and stop_codon.

## Aim

Fully support serialization of GeneModels in GTF format.

## Implementation

Multiple functions for processing GTF coding sequence features were added.
Some functions in `dae.utils.regions` package were fixed. 

Detailed testing of GTF coding sequence features supported.

### DIFF
--- a/dae/dae/utils/regions.py
+++ b/dae/dae/utils/regions.py
@@ -99,7 +99,7 @@ def get_chromosome_length_tabix(
                 left = pos
             else:
                 right = pos
-        return right
+        return right  # noqa: TRY300
 
     except ValueError as ex:
         logger.warning(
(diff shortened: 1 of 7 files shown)


dae/dae/utils/regions.py
```diff
@@ -424,7 +423,7 @@ def collapse_no_chrom(
 
 
 def total_length(regions: list[BedRegion]) -> int:
-    return sum(len(regions) for x in regions)
+    return sum(len(x) for x in regions)
```
iordanivanov, line 426: Does this function have a unit test?

iordanivanov (approved):

dae/dae/utils/regions.py
lchorbadjiev: No. We need to add unit tests for most of the region utils.

PR #681, opened 2024-07-15
Pheno groups

IvoTod:
## Background
Phenotype groups were an old disabled feature that we did not need to use until now.

## Aim
Reimplement and update the phenotype groups code.

## Implementation
Restored and added new missing methods to phenotype groups and updated the fixtures. Families were removed as they are not used and instrument merging was changed to instantly error when the same measure is found in one of the child phenotype studies. Also fixes an issue with the pheno import discovered when updating said fixtures where running the import without regressions would crash.


dae/dae/configuration/schemas/phenotype_data.py
```diff
@@ -37,3 +36,13 @@
     "phenotype_data": {"type": "dict", "schema": phenotype_data_schema},
     **regression_conf_schema,
 }
+
+groups_file_schema = {
+    "groups": {"type": "list", "schema": {
```
lchorbadjiev, line 41: Probably it will be better to name it `pheno_groups`??

dae/dae/gpf_instance/gpf_instance.py
```diff
@@ -194,14 +194,28 @@ def _pheno_registry(self) -> PhenoRegistry:
         pheno_configs = GPFConfigParser.collect_directory_configs(
             pheno_data_dir,
         )
+        groups_config: Path | None = None
```
lchorbadjiev, line 197: `pheno_groups_config`??

lchorbadjiev (approved):
Rename phenotype data `groups` into `pheno_groups` where appropriate.

PR #689, opened 2024-07-25
Pheno experiments

IvoTod:
## Background
When attempting to import phenotype data using workers on multiple nodes, the import was broken and a few adjustments had to be made

## Implementation
The temporary DB file copy is now made in the output directory instead of a tempfile location. This allows workers on different nodes to be able to find the file when properly shared. Also a fix has been added for the `--force` flag which didn't propagate to the DASK executor.


lchorbadjiev (approved):
looks OK to me

PR #697, opened 2024-08-05
Update classification parallelization

IvoTod:
## Background
The pheno import measure classification used a thread pool which was always with only 4 workers and has not been updated for quite some time.

## Aim
Update the pheno import classification step to use a DASK task graph instead.

## Implementation
After changing to a DASK task graph, the previous parallelization turned out to be poorly implemented, due to workers waiting out one another and never doing actual asynchronous work. This introduced an issue with DB access in the workers. Another copying step for the database has been added right after importing all of the necessary measure data for classification. Due to this the `build_pheno_common` function had to be split in two parts for loading the data table and building the measure itself.
A new flag `--skip-pheno-common` was also added.


lchorbadjiev (approved):
LGTM

PR #700, opened 2024-08-07
Vep package

IvoTod:
## Background
The VEP annotation needs to be packaged into its own conda package.

## Aim
Create versioneer for the VEP annotator package.

## Implementation
Due to the way files were organized, the external demo annotators were split from the VEP annotators package.


lchorbadjiev (approved):
LGTM

Probably we should separate running VEP annotator plugin tests into a separate build. But this is something we should do next week.

PR #709, opened 2024-08-29
Parquet variants region fix

iordanivanov:


lchorbadjiev (approved):
LGTM

Please add a simple parquet storage test to import a study and query it.

PR #729, opened 2024-09-30
refactor wdae pheno browser tests

IvoStefanov:
## Aim
Refactor pheno browser wdae tests


iordanivanov (changes requested):
Please use pytest.approx for comparison of float values in tests

iordanivanov (approved):

PR #731, opened 2024-10-01
refactor wdae pheno tool tests

IvoStefanov:
## Aim
Refactor wdae pheno tool tests


iordanivanov (approved):
Again, please use pytest.approx when comparing float values. I have approved it, but fix this before merging

PR #733, opened 2024-10-04
Fix handling of measure names with dot

IvoTod:
## Background
Running the pheno import on a new study revealed an issue with the handling of measure names containing dots.
By design, it's not intended to have measure names with dots in them, but we decided it should be changed as it will most likely not introduce a new bug.

## Aim
Update measure name handling.

## Implementation

When getting the measure and instrument name from the measure id, `.split` was used and the output was being unpacked into 2 variables. These cases now have a `maxsplit` of 1 and should handle measure names with dots.


lchorbadjiev (approved):
Consider adding a unit test before merging

PR #734, opened 2024-10-04
wdae tests: refactor gene scores

joankosev:


iordanivanov (changes requested):
Other than the ``pytest.approx`` comment, I'd like if the commit was renamed as "refactor gene scores" is misleading. Perhaps "refactor gene scores unit tests".

iordanivanov (approved):

PR #739, opened 2024-10-10
Speed up tests with wrong instance

IvoStefanov:
## Background
Updated wdae tests were slow because they used t4c8 wgpf instance that is created anew in every test.

## Aim
Switch the the wgpf instance to a session wide t4c8 instance that is created once for all tests.

### DIFF
--- a/wdae/wdae/pheno_browser_api/tests/test_pheno_browser_api.py
+++ b/wdae/wdae/pheno_browser_api/tests/test_pheno_browser_api.py
@@ -50,7 +50,7 @@ def test_pheno_browser_api_permissions(
     url: str,
     method: str,
     body: dict,
-    t4c8_wgpf: WGPFInstance,  # noqa: ARG001
+    t4c8_wgpf_instance: WGPFInstance,  # noqa: ARG001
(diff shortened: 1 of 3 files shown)


wdae/wdae/pheno_browser_api/tests/test_pheno_browser_api.py
```diff
@@ -276,19 +278,20 @@ def test_measure_details(
 
     assert response.status_code == 200
 
-    print(response.data)
-    assert response.data["instrument_name"] == "i1"
-    assert response.data["measure_name"] == "age"
-    assert response.data["measure_type"] == "continuous"
-    assert response.data["values_domain"] == [
+    data = response.json()
+    print(data)
```
iordanivanov, line 282: Forgotten ``print``

iordanivanov (changes requested):
I would prefer if you rename the commits' messages to something akin to "use session-scoped t4c8 instance in (...)" since "speed up" is vague and non-descriptive

iordanivanov (approved):

PR #741, opened 2024-10-17
Remove gene sets download link

IvoStefanov:
## Background
Gene sets download links were previously sent with the gene sets query. 
Updates on the gene sets api made these links invalid.

## Aim
Remove invalid download links from the gene sets view.
Link will now be created in GPFJS.


iordanivanov (approved):
The message of the second commit is confusing, since syntactically there isn't much difference between Python dictionaries and JSON. Rather, the way ``geneSetsTypes`` seems to have changed is that it mirrors the structure of objects used in the frontend.

PR #758, opened 2024-11-13
Update gene scores histogram api

IvoStefanov:
## Background
Gene scores api is not providing categorical histogram data correctly. Thus using gene scores with such histograms breaks gpf.

## Aim
Create api that gives the whole histogram configuration for all gene scores without manipulating and transforming the data.

## Implementation
Removed partitions api since gpfjs has no use of it.
Create histogram api that provides full unedited histogram information.

## Related issues
The new api is to be used in conjunction with gpfjs branch `add-categorical-histograms-support`. Its purpose is to introduce drawing of interactable categorical histograms in gene scores.


dae/dae/gene_scores/gene_scores.py
```diff
@@ -158,20 +158,28 @@ def get_genes(
...
+                score_max = float("inf")
+
+            index = np.logical_and(
+                df.to_numpy() >= score_min,
+                df.to_numpy() <= score_max)
+            index = np.logical_and(index, df.notna())
+            genes = score_value_df[index].gene
+        else:
+            genes = score_value_df.loc[
+                score_value_df["gene-score"].isin([float(v) for v in values])
```
iordanivanov, line 181: Why are values being cast to float if we are dealing with a categorical gene score (according to the commit message)?

iordanivanov, line 181: Additionally, why is the `score_value_df` being accessed with `"gene-score"`? Is that how the column containing the gene score is always called?

IvoStefanov: "gene-score" should have been the used gene score id. There was a misconception because the code was tested with a gene score with id "gene-score".

IvoStefanov: For now gene score values are numerical. But the frontend uses universally strings because in the future both string and numerical values will be supported. A conversion is needed when searching strings.

iordanivanov (approved):

PR #759, opened 2024-11-14
Refactor permission functions

IvoStefanov:
## Aim
Refactor and bugfix permission functions

## Implementation
Remove unused functions and replace checks for dataset permissions with permitted_datasets method.


IvoTod (changes requested):
The 2nd refactor commit should be split into two separate ones.

IvoTod (approved):

PR #763, opened 2024-11-21
Refactor dae variant utils tests

joankosev:


iordanivanov (commented):
Perhaps the ``get_locus_ploidy`` function should throw an exception if `Sex.F` is provided for a `Y` chromosome? Additionally, consider if the function should throw an exception for `Sex.U`?

iordanivanov (approved):

PR #858, opened 2026-05-19
ci: fetch gain wheel from newest build that has it (skip docs-only greens)

lchorbadjiev:
## Problem

The root `Jenkinsfile` **"Fetch gain wheel"** stage copied `dist/core/*.whl` from `iossifovlab/gain/master` with `selector: lastSuccessful()`.

gain's CI **skips wheel building on docs-only commits**, but the build still **SUCCEEDS** — its only artifact is the docs tarball. So when the newest green `gain/master` build is docs-only, `lastSuccessful()` resolves to a wheel-less build and `copyArtifacts(filter: 'dist/core/*.whl', ...)` matches nothing. This fails **every gpf build on every branch, including master** — fast (~30 s) at "Fetch gain wheel" — until a non-docs gain build supersedes it.

Observed 2026-05-19: `gain/master` [#209](https://nemo.seqpipe.org/job/iossifovlab/job/gain/job/master/209/) (a release-notes docs-only commit) broke gpf builds, e.g. [gpf fix/e2e-db-healthcheck-start-period #5](https://nemo.seqpipe.org/job/iossifovlab/job/gpf/job/fix%252Fe2e-db-healthcheck-start-period/5/): `ERROR: Failed to copy artifacts from iossifovlab/gain/master with filter: dist/core/*.whl`.

## Fix

Query `gain/master`'s Jenkins API for the **newest SUCCESS build whose artifacts actually include `dist/core/*.whl`**, and `copyArtifacts(selector: specific(<that build>))`. Docs-only greens are skipped over instead of breaking the copy.

- `curl` + `python3` only — both always present on the CI agents; **no new Jenkins plugin**.
- `JENKINS_URL` is exported into `sh` by Jenkins.
- Fails loudly with a clear message if *no* recent gain green carries the wheel.

## Notes / test plan

- No local Jenkins declarative/Groovy linter exists in the repo, so this is verified by structure (brace-balanced, minimal diff) only — real proof is a green gpf build.
- [ ] gpf build green on this branch (the "Fetch gain wheel" stage logs `gain-core wheel from gain/master #<n>` and the wheel lands in `dist/gain/`)
- [ ] A subsequent docs-only `gain/master` build does **not** break gpf builds

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -328,7 +328,11 @@ pipeline {
                                 script: '''
                                     set -euo pipefail
                                     url="${JENKINS_URL}job/iossifovlab/job/gain/job/master/api/json?tree=builds[number,result,artifacts[relativePath]]{0,40}"
-                                    curl -fsS "$url" | python3 - <<'PY'
+                                    # -g/--globoff: the Jenkins tree
+                                    # query contains [ ] { } which
+                                    # curl otherwise parses as URL
+                                    # globbing ("bad range in URL").
+                                    curl -fsS -g "$url" | python3 - <<'PY'
 import sys, json
 for b in json.load(sys.stdin).get("builds", []):
     if b.get("result") != "SUCCESS":
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -293,36 +293,72 @@ pipeline {
                 }
 
                 stage('Fetch gain wheel') {
-                    // Pull the gain-core wheel from gain's last
-                    // successful master build into dist/gain/.
+                    // Pull the gain-core wheel into dist/gain/.
                     // Consumed by web_api/Dockerfile.production (the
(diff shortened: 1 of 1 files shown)


lchorbadjiev:
Abandoning the gpf-side approach. Build #1 hit a curl URL-globbing bug (fixed with -g); build #2 then showed the Jenkins controller returns 403 Forbidden to the agent's anonymous api/json request, so a gpf-side REST lookup would need an authenticated Jenkins API token. Per discussion we're pivoting to the gain-side invariant instead: every green iossifovlab/gain/master build will carry dist/core/*.whl (even docs-only ones), which makes gpf's existing 'selector: lastSuccessful()' correct again with no HTTP/auth. gpf master is unchanged (still plain lastSuccessful()). Superseded by a forthcoming iossifovlab/gain PR.

Issue #796, opened 2025-01-28
[EPIC] Reorganization of phenotype data


lchorbadjiev:
### Phenotype Storage Configuration
We are going to model the phenotype storage configuration after the genotype storage configuration in the GPF instance:
```yaml
phenotype_storage:

  default: pheno_storage

  storages:
  - id: pheno_storage
    base_dir: '%(wd)s/phenotype_storage'
```

- Phenotype storage layout:

```
base_dir/
    study1_id/
        study1_id.db
    study2_id/
        study2_id.db
```

### Phenotype Study and Group Configurations
The GPF instance configuration contains a property `phenotype_data` that defines where the configuration of phenotype studies and groups should go:

```yaml
phenotype_data:
  dir: '<path>'
```

If this property is not defined, it defaults to `<gpf_instance dir>/pheno`.

The default structure of the phenotype data configuration directory is as follows:
```
<phenotype_data dir>/
    study1_id/
        study1_id.yaml
    study2_id/
        study2_id.yaml
```

### Phenotype Study Configuration
The phenotype study configuration should contain which storage holds the phenotype databases for the study.
```yaml
id: study1_id
phenotype_storage:
.   id: pheno_storage
.   db: study_id/study_id.db
```

### Default phenotype storage
When the GPF instance has no defined `phenotype_storage` section, it will have the default phenotype storage that is equivalent to the following:
```yaml
phenotype_storage:
    default: default_pheno_storage

    storages:
    - id: default_pheno_storage
      base_dir: '<phenotype_data dir>'
```

In that case, the phenotype data configuration directory will contain phenotype data databases alongside the study configurations:
```
<phenotype_data dir>/
    study1_id/
        study1_id.yaml
        stuty1_id.db
    study2_id/
        study2_id.yaml
        study2_id.db
```

### Phenotype Storage Registry

This class creates and registers all phenotype storages. It accepts a phenotype storage configuration, creates the phenotype storage, and registers it.

* Model the `PhenotypeStorageRegistry` after the `GenotypeStorageRegistry` (https://github.com/iossifovlab/gpf/blob/master/dae/dae/genotype_storage/genotype_storage_registry.py).

### Phenotype Storage
The primary function of phenotype storage is to create phenotype studies.

- Phenotype storage: model it after genotype storage (https://github.com/iossifovlab/gpf/blob/master/dae/dae/genotype_storage/genotype_storage.py)
```python
class PhenotypeStorage(abc.ABC):

[...]

PR #960, opened 2026-07-01
Resolve pheno image dir from instance config when dae_config is absent

lchorbadjiev:
## Background

`get_pheno_db_dir()` and `get_pheno_browser_images_dir()` are called in two modes:
with an already-built `dae_config`, and standalone with `dae_config=None` (e.g. the
pheno-browser image path helpers). In the `None` path, the directory was hard-coded to
`$DAE_DB_DIR/pheno` (and thus `$DAE_DB_DIR/pheno/images`), ignoring any
`phenotype_data.dir` configured in the instance. On instances where phenotype data lives
somewhere other than `$DAE_DB_DIR/pheno`, image paths resolved to the wrong place.

## Aim

Make the `dae_config=None` path resolve phenotype image directories the same way the
configured path does — by reading the instance configuration — so pheno-browser image
paths follow the configured `phenotype_data.dir` instead of a hard-coded sibling
directory.

## Implementation

- `get_pheno_db_dir(None)` now loads the instance config from
  `$DAE_DB_DIR/gpf_instance.yaml` via `GPFInstance.build_gpf_config(...)` and resolves the
  configured `phenotype_data.dir` (falling back to `<conf_dir>/pheno`), instead of
  returning `$DAE_DB_DIR/pheno` unconditionally. Raises a clear `ValueError` if the config
  still can't be built (mis-set `DAE_DB_DIR`).
- Promoted `GPFInstance._build_gpf_config` → public `GPFInstance.build_gpf_config` so the
  pheno layer (core) can build a config without reaching into a private staticmethod; the
  `web_api` `WGPFInstance` caller drops its `# noqa: SLF001`.
- The `gpf.gpf_instance` import in `pheno_data.py` is done lazily inside the function to
  avoid a core-internal import cycle (`pheno` ← `gpf_instance` ← … ← `pheno`).

Checks: `ruff` clean; `mypy gpf` (191 files) and `mypy gpf_web` (250 files) both clean.

### Notes for review
- Behaviour change: callers relying on the old `$DAE_DB_DIR/pheno` fallback for a
  `None` config will now get the configured dir. Worth confirming no caller depended on the
  old hard-coded path.
- No regression test is included yet — happy to add one exercising `get_pheno_db_dir(None)`
  against a fixture instance with a non-default `phenotype_data.dir` if you'd like it in
  this PR.

## Related issues

Related to iossifovlab/gpf#796 ([EPIC] Reorganization of phenotype data). Does not close it.

### DIFF
--- a/core/gpf/configuration/gpf_instance_config.py
+++ b/core/gpf/configuration/gpf_instance_config.py
@@ -1,0 +1,51 @@
+"""Discovery and loading of the top-level GPF instance configuration.
+
+This lives in the low-level ``configuration`` package (rather than on
+``GPFInstance``) so that other core modules — notably ``gpf.pheno`` — can
+resolve the instance config without importing ``gpf.gpf_instance``, which
+would create an import cycle (``gpf_instance`` imports ``pheno``).
+"""
+from __future__ import annotations
+
+import os
+from pathlib import Path
+
+from box import Box
+from gain.utils.fs_utils import find_directory_with_a_file
+
+from gpf.configuration.gpf_config_parser import GPFConfigParser
+from gpf.configuration.schemas.dae_conf import dae_conf_schema
+
+
+def build_gpf_config(
+    config_filename: str | Path | None = None,
+) -> tuple[Box, Path, Path]:
+    """Build GPF config from a config file or environment variable.
+
+    When ``config_filename`` is ``None``, the instance is discovered via
+    ``GPF_CONF_DIR``, then ``DAE_DB_DIR``, then a walk of the current
+    directory and its parents for a ``gpf_instance.yaml``.
+    """
+    dae_dir: Path | None
+    if config_filename is not None:
+        config_filename = Path(config_filename)
+        dae_dir = config_filename.parent
+    else:
+        if os.environ.get("GPF_CONF_DIR"):
+            dae_dir = Path(os.environ["GPF_CONF_DIR"])
+            config_filename = Path(dae_dir) / "gpf_instance.yaml"
+        elif os.environ.get("DAE_DB_DIR"):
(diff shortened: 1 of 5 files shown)


lchorbadjiev:
Addressed the review findings in `2830b66`:

- **#1 (blocker) — existing tests failed CI.** `test_get_pheno_db_dir` / `test_get_pheno_browser_images_dir` exercise the *undiscoverable-instance* fallback; they now pass again (and clear `GPF_CONF_DIR` for isolation). Added two regression tests that assert the `None`-arg path matches the explicit-config resolution for `phenotype_data.dir` and `phenotype_images`.
- **#2 (blocker) — import-time crash.** `get_pheno_db_dir` / `get_pheno_browser_images_dir` now **fall back to `<DAE_DB_DIR>/pheno`** instead of raising when no instance is discoverable, so settings import (`settings.py:29`, `wgpf_settings.py:37`) is safe again.
- **#3 — `None` branch ignored `phenotype_images`/`cache_path`.** The discovered config is now routed through the same logic as the explicit-config branch, so all three keys are honored consistently.
- **#4 — re-implemented discovery.** Now calls `GPFInstance.build_gpf_config()` with no args, using its real ladder (`GPF_CONF_DIR` → `DAE_DB_DIR` → parent-walk) via a small `_discover_dae_config()` helper — no more CWD-relative bare filename.
- **#5 — dead `ValueError` branch.** Removed; replaced by the graceful fallback.
- **#6 — layering.** The lazy import stays (it genuinely breaks the `gpf_instance` ↔ `pheno_data` import cycle) and is now documented as such, with the complexity behind `_discover_dae_config()`.

Checks: `ruff` clean (no new findings), `mypy gpf` clean (191 files), full `tests/small/pheno/` suite green (222 passed).

Issue #965, opened 2026-07-07
Pheno browser: hide/disable Download button per pheno_browser_download feature flag (SPA wiring)


lchorbadjiev:
## Summary

Wire the Angular SPA to the `/api/v3/instance/features` endpoint so the phenotype-measures **Download** control reflects the `pheno_browser_download` feature flag. Today the flag is backend-only: turning it off hardens the API but leaves a visible, inert Download button in the UI.

## Background

The `feature-flags` work (merged in gpf commit `78bd95e`) added:

- A flag registry exposed at `GET /api/v3/instance/features` (returns a `name -> enabled` map, e.g. `{"pheno_browser_download": true}`).
- Backend gating of `GET/HEAD /api/v3/pheno_browser/download` via `FeatureFlagMixin`, which returns **404** when the flag is disabled (before auth, so the endpoint looks absent).

The backend review findings (registry merge-over-defaults, reset hook, tests, type hint, ETag) are resolved. The remaining item is the frontend, which was deliberately scoped out of the backend pass.

## Problem

The SPA never reads `/features`. `pheno-browser.component.ts` always renders the Download control and, on click, calls `phenoBrowserService.validateMeasureDownload()` which issues a `HEAD` to `pheno_browser/download` and switches on the status:

```
if (validity.status === 200)  { window.open(downloadLink); }
else if (validity.status === 204) { /* "select more than 0" */ }
else if (validity.status === 413) { /* "too many measures" */ }
// no branch for 404
```

With the flag off, the `HEAD` returns **404**, which matches none of the branches, so the click **silently does nothing** — no download, no error message. The button is present but dead.

## Proposed change

1. Fetch `/api/v3/instance/features` (at app/config load, or when the pheno browser initializes) and expose the flag map to the pheno-browser component.
2. Hide or disable the Download button when `pheno_browser_download` is `false`.
3. Defensively handle a 404 from `validateMeasureDownload()` (e.g. a clear "download is disabled" state) in case the flag flips between load and click.

## Affected files

- `web_ui/src/app/pheno-browser/pheno-browser.component.ts` — `downloadMeasures()` handler and template.
- `web_ui/src/app/pheno-browser/pheno-browser.service.ts` — `validateMeasureDownload()` / download helpers; likely a new `getFeatures()` call.
- Wherever instance-level config is loaded at startup (candidate home for the `/features` fetch).

## Acceptance

[...]

PR #966, opened 2026-07-07
Harden pheno-download feature flags: merge-over-defaults, reset hook, tests

lchorbadjiev:
## Background

The `feature-flags` merge (`78bd95e`) added a feature-flag facility to the GPF web API and used it to gate the phenotype-measures CSV download (`GET/HEAD /api/v3/pheno_browser/download`) behind the `pheno_browser_download` flag, plus a `GET /api/v3/instance/features` discovery endpoint. A code review of that merge surfaced a set of backend findings; this PR addresses them. It is backend-only and backwards compatible (the flag still defaults on).

## Aim

Harden the feature-flag facility so it is safe to override per-deployment, testable, and covered:

- **Overrides can't accidentally disable a known flag.** Previously `settings.FEATURE_FLAGS` was the whole registry and unknown keys read as disabled, so a deployment that set its own `FEATURE_FLAGS` without re-listing `pheno_browser_download` would silently 404 the download.
- **The gate is testable.** The process-wide singleton cached flags for the life of the process with no reset hook, so `override_settings(FEATURE_FLAGS=...)` had no effect — which is also why the behaviour was untested.

## Implementation

- **Merge-over-defaults.** `DEFAULT_FEATURE_FLAGS` in `gpf_instance/feature_flags.py` is now the canonical in-code registry; `get_feature_flags()` computes `{**DEFAULT_FEATURE_FLAGS, **settings.FEATURE_FLAGS}`. `default_settings.FEATURE_FLAGS` becomes an empty per-deployment override layer, so a deployment only lists the flags it wants to change.
- **`reset_feature_flags()`** drops the cached singleton under the existing lock; tests use a `reset_flags` fixture that resets before/after each toggling test.
- **Tests.** New `gpf_instance/tests/test_features_api.py` covers `/features` (map, overrides, merged defaults, ETag). New tests in `pheno_browser_api/tests/test_pheno_browser_api.py` assert a 404 on `GET` and `HEAD` of the download when the flag is off, and that an override omitting the flag leaves the download enabled — the existing enabled-path tests still pass in the same run.
- **`FeatureFlagMixin.dispatch`** now annotates `request` as `django.http.HttpRequest` (it runs before DRF wraps the request) and the unused DRF `Request` import is dropped. The deferred `from django.conf import settings` is hoisted to module level (it's a lazy proxy — nothing is evaluated at import), removing the `import-outside-toplevel` suppression.
- **ETag** added to the `features` view, matching the `version` endpoint.

Verification: `29 passed` across the three affected test files run together (no cross-file singleton leakage); `ruff` clean, `mypy` clean, `pylint` 10.00/10 on the changed files.

## Related issues

- Frontend follow-up (hide/disable the Download button when the flag is off): #965

### DIFF
--- a/web_api/gpf_web/conftest.py
+++ b/web_api/gpf_web/conftest.py
@@ -3,7 +3,7 @@
 
 import logging
 import pathlib
-from collections.abc import Sequence
+from collections.abc import Iterator, Sequence
 from datetime import timedelta
 from typing import cast
 
@@ -22,6 +22,7 @@ from gain.genomic_resources.testing import (
     build_inmemory_test_repository,
     convert_to_tab_separated,
 )
+from gpf_instance.feature_flags import reset_feature_flags
 from gpf_instance.gpf_instance import (
     WGPFInstance,
     get_wgpf_instance,
@@ -43,6 +44,18 @@ from gpf.gpf_instance.gpf_instance import GPFInstance
 from gpf.studies.study import GenotypeData
 
 logger = logging.getLogger(__name__)
+
+
+@pytest.fixture
+def reset_flags() -> Iterator[None]:
+    """Drop the cached feature-flags singleton around a test.
+
+    Reset before and after so ``override_settings(FEATURE_FLAGS=...)`` takes
+    effect and no toggled state leaks into other tests.
+    """
+    reset_feature_flags()
+    yield
+    reset_feature_flags()
 
 
 @pytest.fixture
--- a/web_api/gpf_web/gpf_instance/feature_flags.py
+++ b/web_api/gpf_web/gpf_instance/feature_flags.py
@@ -26,7 +26,9 @@ class FeatureFlags:
     """Singleton registry of web-API feature flags."""
 
     def __init__(self, flags: dict[str, bool]) -> None:
-        self._flags = dict(flags)
+        # Coerce to bool so a mistyped settings value (e.g. the string
+        # "false") can't leak a non-bool through is_enabled().
+        self._flags = {name: bool(value) for name, value in flags.items()}
 
     def is_enabled(self, name: str) -> bool:
         """Return True if the named feature is enabled."""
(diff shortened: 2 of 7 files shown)


lchorbadjiev:
## Code review findings

An independent review pass on this branch. The change is mergeable as-is — the core merge-over-defaults fix is correct and test-covered — but a few points are worth tightening. Filing them here; addressing 1, 2, 3, 5 in a follow-up commit on this branch.

**1. Unknown override keys pass through silently (should-fix).** `flags = {**DEFAULT_FEATURE_FLAGS, **overrides}` accepts any key from `settings.FEATURE_FLAGS`, even one not in the canonical registry. A typo like `FEATURE_FLAGS = {"pheno_browser_downlod": False}` (a) does *not* disable the real `pheno_browser_download`, and (b) adds a bogus flag that `GET /api/v3/instance/features` then advertises. This is the mirror image of the footgun the PR set out to fix. → Drop override keys not in `DEFAULT_FEATURE_FLAGS` and log a warning.

**2. `is_enabled()` can return a non-bool (nit).** It returns the raw dict value, so a stringly-typed setting like `{"pheno_browser_download": "false"}` returns a truthy string despite the `-> bool` contract. → Coerce values to `bool` at the `FeatureFlags` boundary.

**3. `reset_flags` fixture duplicated** verbatim in `gpf_instance/tests/test_features_api.py` and `pheno_browser_api/tests/test_pheno_browser_api.py`. → Hoist to the shared `gpf_web/conftest.py`.

**4. Gated `HEAD` returns an HTML-bodied 404 (nit, won't fix).** `HttpResponseNotFound()` carries `<h1>Not Found</h1>`; strictly a HEAD body should be empty. This is consistent with *every* Django 404 in the app (routing 404s included), so special-casing method semantics inside a generic gating mixin would be inconsistent and isn't worth it. Noting and declining.

**5. No isolated unit test of the merge/`FeatureFlags` logic (nit).** All behavior is exercised via HTTP. → Add a direct unit test that pins the merge, the coercion (2), and the unknown-key drop (1).

### Verified correct during review
Double-checked-locking + `reset_feature_flags()` race behavior; the hoisted module-level `settings` import (lazy proxy, safe at any app-load order); MRO with `FeatureFlagMixin` first and gating-before-auth (uniform 404, no info leak); the `HttpRequest` type-hint accuracy; `@etag` semantics (flags are process-constant, timestamp bumps on restart); `default_settings.FEATURE_FLAGS = {}` causing no regression (nothing else reads it); and the `reset_flags`/`override_settings` fixture ordering being reliable under xdist.

lchorbadjiev:
Findings addressed in `4ccf3ba12`:

- **1** — `get_feature_flags()` now drops override keys not in `DEFAULT_FEATURE_FLAGS` and logs a warning (a typo'd flag can neither fail to disable the real one nor be advertised via `/features`).
- **2** — `FeatureFlags` coerces values to `bool`.
- **3** — `reset_flags` fixture hoisted to the shared `gpf_web/conftest.py`.
- **5** — new `test_feature_flags.py` with direct unit tests for the merge, coercion, and unknown-key drop.
- **4** — intentionally left as-is (consistent with every Django 404 in the app; not worth coupling the generic mixin to HTTP-method semantics).

Verification: 35 tests green across the affected files; ruff clean, mypy clean, pylint 10.00/10.

PR #974, opened 2026-07-14
web_e2e: switch the CI GRR from the dead NFS mount to https://grr.seqpipe.org + persistent cache

### LINKED ISSUE #973: gpf-web-e2e: switch the GRR from the dead NFS mount to the node-local grr-sync directory GRR
> *This was generated by AI during triage.*

Companion to the umbrella **iossifovlab/gain#269**. Cuts `gpf-web-e2e` from the bind-mounted NFS GRR (`/mnt/cephfs/seqpipe/grr` + `/mnt/cephfs/seqpipe/grr_sfari` — **piglet is decommissioned**) to **node-local `type: directory` GRRs** produced by `grr-sync` (**iossifovlab/grr-sync#20**).

**This supersedes the earlier HTTP+cache scoping of this issue** (never implemented on gpf). See #269 for the reasoning: the whole-file `cache_dir` turns KB tabix range reads into multi-GB downloads; a node-local directory GRR gives byte-range local reads, zero per-build network, and build-time isolation from every remote.

## Changes

**1. `web_e2e/gpf_e2e_instance/grr-definition.yaml` stays exactly as-is** — a `group` of `type: directory` children (`grr_sfari.sync` first so its enrichment backgrounds override matching ids in `grr`, then `grr.sync`), ids unchanged, pointing at `/grr_sfari` and `/grr`. **No `http`, no `cache_dir`, no `read_only` churn.**

**2. Repoint the three compose files** — `web_infra/compose-jenkins.yaml`, `compose-jenkins-split.yaml`, `compose-jenkins-combined.yaml` — changing the `/grr` and `/grr_sfari` mounts from `/mnt/cephfs/...` to the node-local `grr-sync` stable symlinks, **read-only**:
```yaml
- ${GRR_ROOT}/grr:/grr:ro
- ${GRR_ROOT}/grr_sfari:/grr_sfari:ro
```

**3. Add a fail-fast pre-flight** (both suites): before `docker compose up`, assert each required GRR symlink resolves **and** its `.CONTENTS.json.gz` exists; else fail with `"GRR not synced on this node yet"` (prevents the empty-bind-mount footgun).

**4. `web_e2e/README.md`** — update the two documented CSHL bind-mounts to the node-local model.

## Acceptance

Re-enable the job; a green run on an agent carrying fresh `grr-sync` snapshots. Node-local trees mean no cold-download penalty.

## Depends on

- **iossifovlab/grr-sync#20** (produces the node-local trees) — blocker
- **seqpipe/infra#71** (deploys grr-sync on the agents) — blocker

lchorbadjiev:
Closes #973. Companion to iossifovlab/gain#271 (same change for `gain-web-e2e`); part of iossifovlab/gain#269.

`gpf-web-e2e` is currently **disabled**: `instance-import` (base), split-mode `backend` and combined-mode `frontend` all bind-mounted `/mnt/cephfs/seqpipe/grr` + `/mnt/cephfs/seqpipe/grr_sfari`, and piglet — the NFS server behind `/mnt/cephfs` — is decommissioned. A bind-mount of a missing host path yields an *empty* directory, so the GRR read empty.

This points CI at the two http GRRs `grr-sync` mirrors on kowalski — **https://grr.seqpipe.org** (`.CONTENTS.json.gz` byte-identical to `grr.iossifovlab.com`) and **https://grr-sfari.seqpipe.org** — behind a **persistent per-agent cache**.

**Production is untouched.** It keeps the image-baked `web_api/scripts/grr-definition.yaml` (`directory: /grr`) and CSHL's own on-host GRR. Only the e2e instance's definition changes, and that file is bind-mounted from the checkout, not baked into any image — so no Dockerfile change is needed here (this is where gpf differs from gain#271, which had to add `grr-definition-http.yaml` to its image).

## Changes

| File | Change |
|---|---|
| `web_e2e/gpf_e2e_instance/grr-definition.yaml` | still a `group`, same two children in the same order — now `type: http` against seqpipe.org, with `cache_dir: /grr_cache` on the group |
| `web_infra/compose-jenkins.yaml` | `instance-import`: drop the two NFS binds, add `${GRR_CACHE_DIR}:/grr_cache`, run as the agent uid, `HOME=/tmp` |
| `web_infra/compose-jenkins-split.yaml` | same for `backend`, plus `start_period: 3600s` on its healthcheck |
| `web_infra/compose-jenkins-combined.yaml` | same cache mount + `start_period` for `frontend`; it **stays root** (see below) |
| `web_e2e/Jenkinsfile.e2e` | `GRR_CACHE_DIR`/`AGENT_UID`/`AGENT_GID`; `mkdir -p` the cache and the `wdae-logs` bind source; chown the workspace + cache back to the agent uid; job timeout 1h → 4h |
| `web_e2e/README.md`, `web_infra/.gitignore` | document the https + cache model; ignore the local default cache dir |

## Four load-bearing, non-obvious details

- **Child order is preserved.** `grr_sfari` stays first: a group resolves an id from the first child that has it, and grr_sfari's enrichment backgrounds deliberately override the matching ids in `grr`.
- **Child ids stay verbatim.** A group-level `cache_dir` auto-partitions by child repo id (`_get_or_create_cache_proto` does `os.path.join(self.cache_url, proto_id)`), so the cache materialises at `/grr_cache/grr.sync/` and `/grr_cache/grr_sfari.sync/`. A rename would silently orphan the warm cache rather than fail loudly. `grr.sync` is also what gain's CI definition calls the main GRR, so **gpf-web-e2e and gain-web-e2e share one cached copy** of it on the agent.
- **No `read_only: true`.** It exists on `FileRepoDefinition` (what the old `type: directory` children were) but *not* on `HttpRepoDefinition`, and `_RepoDefinitionBase` sets `extra="forbid"` — carrying it over is a hard validation failure.
- **The cache's unit is a whole file, not a byte range.** A resource the suite touches at all is downloaded in full — ~32.5 GB worst case (dbSNP 15 GB, gnomAD genomes 11 GB, MPC 2.4 GB, gnomAD exomes 1.4 GB, ClinVar 1.4 GB, GRCh38 0.8 GB). Hence job timeout **1h → 4h** and healthcheck `start_period` **3600s** on the web tier: a slow cold build must not be converted into a *failed* healthcheck (`dependency failed to start`). A probe failing inside `start_period` burns no retry, so a warm cache pays nothing for the budget. Large downloads need no new retry logic — gain#43 already shipped the aiohttp timeout + whole-file retry with backoff. Both numbers deserve trimming once a real cold run and a real warm run have been measured.

## Non-root, and the one exception

`instance-import` and the split-mode `backend` run as `${AGENT_UID}:${AGENT_GID}`. The cache is long-lived and shared; a root-owned cache tree is one the `jenkins` user can neither prune nor refresh — the EPERM-on-root-owned-artefacts bug class this stack has already hit. Consequences handled in the Jenkinsfile: `HOME=/tmp` (the image has no passwd entry for the agent uid), `wdae-logs` is `mkdir`'d on the host (dockerd creates a *missing* bind source as `root:root`), and the workspace + cache are chown'd back to the agent uid at the start of every build — which is also the only way a now-non-root `instance-import` can `rm -rf` the root-owned import residue earlier builds left in `gpf_e2e_instance/`.

**Combined mode stays root** — the deviation from gain#271, which had no such service. The combined image is the production artefact: its `supervisord.conf` declares `user=root` and hard-exits with `Can't drop privilege as nonroot user` under any other uid (verified against a rebuild of that image's runtime layers), after which `apache2ctl` still has to create `/var/run/apache2` and bind `:80`. Overriding that (a CI-only supervisord conf + a tmpfs over `/run`) would reshape the very artefact combined mode exists to test faithfully. The per-build chown contains the fallout instead, combined mode is opt-in (`STACK_MODE` defaults to `split`), and the non-root `instance-import` warms most of the cache before it ever starts.

## Validation

I cannot run the e2e suite or Jenkins from here, so:

- **The new definition was built through gain's own factory**, not eyeballed: `build_genomic_resource_repository(file_name=...)` returns a `GenomicResourceCachedRepo` with `cache_url == "file:///grr_cache"`, child order `['grr_sfari.sync', 'grr.sync']`, and cache protocols resolving to `<cache_dir>/grr_sfari.sync` and `<cache_dir>/grr.sync` (confirmed on disk against a writable temp `cache_dir`). Adding `read_only: true` to the same dict is **rejected** — `group.children.0.http.read_only: Extra inputs are not permitted [type=extra_forbidden]` — confirming that trap is real. The prod definition still loads unchanged.
- `docker compose config` on **both** overlays interpolates and validates: `user: 1000:1000` on `instance-import` + `backend` (and no `user` on the combined `frontend`), `source: /var/lib/jenkins/grr_cache → /grr_cache` on all three GRR readers, `start_period: 1h0m0s` on the web tier, and **no `/mnt/cephfs` bind anywhere**.
- The `Prepare workspace` shell body was executed verbatim against a simulated workspace seeded with root-owned residue: it reclaims the import tree and the cache, wipes the reports, and creates `wdae-logs` host-side — after which a non-root `rm -rf` of the previously-root-owned tree succeeds.
- `web_e2e/Jenkinsfile.e2e` passes the live controller's declarative linter.

**Not verified** (needs a real run): the actual cold-download volume and whether 4h / 3600s are enough; that nothing else in the backend image needs root at runtime; and that `${HOME}/grr_cache` is writable by the Jenkins user on every `!dory` agent. The job also has to be **re-enabled** in Jenkins — a green cold run followed by a fast warm run is the acceptance test, per #973.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/web_e2e/Jenkinsfile.e2e
+++ b/web_e2e/Jenkinsfile.e2e
@@ -238,7 +238,18 @@ pipeline {
                 // creates a *missing* bind source as root:root — a
                 // dir the non-root instance-import / backend could
                 // not write into.
+                //
+                // GRR_CACHE_DIR / AGENT_UID / AGENT_GID are re-derived from
+                // the shell rather than trusted from the environment{}
+                // block — see the `Compose env` stage for why. Here it also
+                // matters for correctness of the chown: a `chown -R ":"`
+                // (empty uid:gid) is an error, and `mkdir -p ""` is too.
                 sh '''
+                    : "${GRR_CACHE_DIR:=$HOME/grr_cache}"
+                    AGENT_UID="$(id -u)"
+                    AGENT_GID="$(id -g)"
+                    export GRR_CACHE_DIR AGENT_UID AGENT_GID
+
                     mkdir -p "$GRR_CACHE_DIR"
 
                     docker run --rm \
@@ -275,6 +286,99 @@ pipeline {
                                       depth: 1]],
                     ])
                 }
+            }
+        }
+
+        // Make the compose stack's view of GRR_CACHE_DIR / AGENT_UID /
+        // AGENT_GID deterministic, then PROVE it before anything starts.
+        //
+        // gain-web-e2e #1042 (identical compose pattern) declared all three
+        // in the pipeline `environment{}` block and `docker compose` still
+        // did not see them: `docker inspect` on the live container showed
+        // the bind source as <workspace>/web_infra/grr_cache — the
+        // `${GRR_CACHE_DIR:-./grr_cache}` FALLBACK, i.e. inside the
+        // workspace, wiped by the next build's cleanup — and User=0:0, the
+        // `${AGENT_UID:-0}` fallback. Both designs, the persistent cache and
+        // the non-root ownership, were completely inert, and nothing failed.
+        //
+        // Rather than chase Jenkins' env-export semantics, pin it:
+        //
+        //   1. Re-derive the values HERE, in the shell, from `$HOME` and
+        //      `id -u`/`id -g` — the shell's own view is authoritative and
+        //      cannot be half-exported.
+        //   2. Write them to web_infra/.env — the compose PROJECT directory
+        //      (the dir of the first `-f` file), which is where Compose
+        //      auto-loads a .env from.
+        //   3. Pass `--env-file web_infra/.env` explicitly on every compose
+        //      call anyway, so nothing depends on which directory Compose
+        //      decides the project dir is.
+        //   4. `set -a; . ./web_infra/.env` before each compose call, because
+        //      --env-file is NOT enough on its own: the process environment
+        //      OUTRANKS --env-file, so a var that is present-but-EMPTY in the
+        //      shell still wins and still selects the fallback (verified
+        //      against docker compose). Sourcing makes the shell agree with
+        //      the file, which closes that hole.
+        //
+        // The ${VAR:-default} fallbacks stay in the compose YAML as a
+        // last-resort guard for local `docker compose up` outside CI.
+        stage('Compose env') {
+            steps {
+                sh '''
+                    set -eu
+
+                    : "${GRR_CACHE_DIR:=$HOME/grr_cache}"
+                    mkdir -p "$GRR_CACHE_DIR"
+
+# Heredoc body is at column 0 on purpose: compose's dotenv parser takes
+# the line verbatim, and a leading-space "  KEY=value" is not reliably
+# the same thing as "KEY=value".
+cat > web_infra/.env <<EOF
+GRR_CACHE_DIR=$GRR_CACHE_DIR
+AGENT_UID=$(id -u)
+AGENT_GID=$(id -g)
+EOF
+
+                    echo "--- web_infra/.env ---"
+                    cat web_infra/.env
+                '''
+
+                // The teeth. A silent fallback to a workspace-local cache
+                // running as root is EXACTLY the bug above; it must never
+                // pass quietly again. This asserts on the RESOLVED config
+                // (`docker compose config`), not on what we believe we
+                // exported, and fails the build with expected-vs-actual.
+                //
+                // BOTH overlays are checked on every build, not just the
+                // STACK_MODE one: split is the default, so a regression in
+                // the combined overlay would otherwise sit unnoticed until
+                // someone opted into combined mode.
+                sh '''
+                    set -eu
+                    set -a; . ./web_infra/.env; set +a
+
+                    echo "########## split mode ##########"
+                    EXPECT_GRR_SERVICES="instance-import backend" \
+                    ./web_infra/verify-compose-env.sh \
+                        web_infra/.env \
+                        web_infra/compose-jenkins.yaml \
+                        web_infra/compose-jenkins-split.yaml
+
+                    echo
+                    echo "########## combined mode ##########"
+                    # `frontend` is the documented root exception: the
+                    # combined image IS the production artefact, and its
+                    # supervisord.conf declares user=root and hard-exits
+                    # ("Can't drop privilege as nonroot user") for anyone
+                    # else. It still gets the cache bind + GRR_DEFINITION_FILE;
+                    # only the uid leg is waived, and the `Prepare workspace`
+                    # chown reclaims whatever it leaves root-owned.
+                    EXPECT_GRR_SERVICES="instance-import frontend" \
+                    ROOT_OK_SERVICES="frontend" \
+                    ./web_infra/verify-compose-env.sh \
+                        web_infra/.env \
+                        web_infra/compose-jenkins.yaml \
+                        web_infra/compose-jenkins-combined.yaml
+                '''
             }
         }
 
@@ -503,7 +607,9 @@ pipeline {
             // committed package-lock.json and the test sources).
             steps {
                 sh '''
+                    set -a; . ./web_infra/.env; set +a
                     docker compose \
+                        --env-file web_infra/.env \
                         -p "$COMPOSE_PROJECT" \
                         -f web_infra/compose-jenkins.yaml \
                         -f $COMPOSE_OVERLAY \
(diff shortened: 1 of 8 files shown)


lchorbadjiev:
## Fix: the persistence + non-root designs in this PR were **inert**

Proven on **gain-web-e2e #1042** (identical compose pattern). `docker inspect` on the live container:

| | expected | **actual on #1042** |
|---|---|---|
| `/grr_cache` bind source | `${HOME}/grr_cache` | **`<workspace>/web_infra/grr_cache`** |
| container `User` | agent uid | **`0:0`** |

Compose took the `${GRR_CACHE_DIR:-./grr_cache}` and `${AGENT_UID:-0}:${AGENT_GID:-0}` **fallbacks** even though `environment{}` sets all three. The "persistent" cache landed in the workspace (wiped by `cleanWs`), and everything ran as root. Nothing failed.

### The fix — deterministic, not archaeological

1. New **`Compose env`** stage re-derives the values **in the shell** (`$HOME`, `id -u`, `id -g`) and writes **`web_infra/.env`** (the compose project dir). **`.env` was NOT gitignored in gpf** — added to `web_infra/.gitignore` (gain already had it).
2. **`--env-file web_infra/.env`** on every `docker compose` call (9 invocations: build / up / wait / run / logs / down).
3. Each call also **sources** that `.env` first.

**Why (3) is load-bearing** — `.env` + `--env-file` alone would *not* have fixed this. The process env **outranks** `--env-file`, so a var that is present-but-**empty** in the shell still wins and still picks the fallback. Verified:

```
CASE: var EMPTY in shell, --env-file given
    user: '0:0'
    source: .../proj/grr_cache        <-- STILL the fallback
CASE: var EMPTY in shell, .env SOURCED + --env-file
    user: 1000:1000
    source: /tmp/real_cache           <-- correct
```

`Prepare workspace` re-derives them the same way, so its `chown -R` can no longer run as an empty `":"`. The `${VAR:-default}` fallbacks stay in the YAML for local `compose up`.

### Verification with teeth — `web_infra/verify-compose-env.sh`

Parses `docker compose config` and **fails the build** unless the `/grr_cache` bind source is exactly `$GRR_CACHE_DIR` (under `$HOME`, **outside** `$WORKSPACE`), every GRR-building service sets `GRR_DEFINITION_FILE`, and the resolved user is non-root — each with expected-vs-actual. Also rejects any surviving `/mnt/cephfs` bind, and **refuses to pass if it matched no service at all** (a check that asserts nothing is how this shipped).

It runs against **both overlays on every build**, not just the `STACK_MODE` one — split is the default, so a combined-mode regression would otherwise sit unnoticed until someone opted in.

Proven against the real failure states: #1042 verbatim (workspace bind + `0:0`) ❌ · missing `GRR_DEFINITION_FILE` ❌ · surviving `/mnt/cephfs` bind ❌ · `GRR_CACHE_DIR` inside `$WORKSPACE` ❌ — all fail loudly.

### Audit: every service that builds a GRR

All three already carried the cache bind + `GRR_DEFINITION_FILE`; the verifier now **holds them to it**. Combined-mode `frontend` stays root (the documented supervisord exception) — declared explicitly via `ROOT_OK_SERVICES=frontend` rather than being skipped, so it still has to satisfy the bind + definition legs.

### Evidence — resolved `docker compose config`

**split** (`compose-jenkins.yaml` + `compose-jenkins-split.yaml`):
```
SERVICE            /grr_cache-BIND-SOURCE   USER         GRR_DEFINITION_FILE
backend            $GRR_CACHE_DIR           1000:1000    /data/grr-definition.yaml
instance-import    $GRR_CACHE_DIR           1000:1000    /data/grr-definition.yaml
db / mail / mysql-data / frontend / e2e-tests   (no GRR)
```

**combined** (`compose-jenkins.yaml` + `compose-jenkins-combined.yaml`):
```
SERVICE            /grr_cache-BIND-SOURCE   USER         GRR_DEFINITION_FILE
frontend           $GRR_CACHE_DIR           0:0 (*)      /data/grr-definition.yaml
instance-import    $GRR_CACHE_DIR           1000:1000    /data/grr-definition.yaml
(*) documented root exception — supervisord hard-exits as non-root
```

No `/mnt/cephfs` bind in any resolved config. Jenkinsfile validated against the live declarative linter; `verify-compose-env.sh` is shellcheck-clean. No Python touched, so ruff/mypy/pylint are unaffected.

**GRR URLs unchanged** — `grr.seqpipe.org` / `grr-sfari.seqpipe.org` stay (kowalski's mirror is fixed).

Not verifiable here: no Jenkins/e2e run. The first CI build on this branch is the real test — and it can no longer pass silently wrong.

PR #975, opened 2026-07-14
Conda env: track pyproject constraints; remove obsolete requirements.txt

lchorbadjiev:
Conda environment hygiene, surfaced while doing the Dependabot django bump (#972).

## 1. `environment.yml` now tracks `pyproject.toml` instead of exact pins

`environment.yml` carried exact `=` pins while the workspace `pyproject.toml` files express floors — and nothing re-derived the former when the latter moved. It had rotted accordingly: **35 of its 43 pins lagged `uv.lock`**, e.g. `numpy=2.2.6` (lock: 2.4.4), `lark=1.2.2` (1.3.1), `scikit-learn=1.5` (1.8.0), `gunicorn=22.0` (25.3.0).

**That rot had teeth.** The pin was `django=5.2.5`, which sits *inside* the `CVE-2026-7666` vulnerable range (`>=5.2, <5.2.15`) that #972 fixes in `uv.lock`. Dependabot only scans `uv.lock`, so it never flagged the conda env — but the README tells developers to build their environment from this file. The exact pin was actively holding local dev environments on a vulnerable Django.

So the fix is not "re-pin to today's versions" (that just re-rots in a month). Instead the file now **mirrors the workspace `[project.dependencies]`** — same floors, same upper bounds. It tracks pyproject by construction and cannot drift the same way.

All **44** runtime deps (including web_api's `mysql` extra) now match their pyproject constraint exactly — verified mechanically, 0 missing, 0 mismatched.

Entries with no pyproject counterpart are grouped under an explicit conda-only heading, with reasons: `htslib`/`samtools`/`bcftools`, `mysql-common`, `jupyter-server-proxy`, `setuptools`, `pip`, `pyopenssl`. All keep their existing pins except `htslib`/`samtools`/`bcftools` — see below.

`python=3.12` is the one deliberate exception to the floors rule, and the header now says so. Every pyproject declares `requires-python = ">=3.12"`, but the project targets 3.12 itself: CI builds and tests on it, and `web_api/Dockerfile.production`'s wheels-only invariant assumes cp312 wheels. A `python>=3.12` floor would let the solver take a newer interpreter and silently diverge this env from what CI tests. The exception is cheap, though — conda's `=` is a prefix match, so `python=3.12` already accepts every 3.12.x patch (the solve picks 3.12.13). It was never exposed to the rot that bit `django=5.2.5`, which froze an exact *patch*.

`htslib`/`samtools`/`bcftools` become floors (`>=1.22`) too. Their old rationale — "the system binaries behind pysam" — was simply wrong: conda's pysam bundles its own htslib and declares **no dependency on the htslib package at all**, and nothing in gpf or gain links against these or shells out to them. They are a developer convenience — daily work on gain/gpf constantly wants these CLIs at hand. That reframing is what makes the floor obviously safe: `1.22` is "recent enough to be useful", not a compatibility contract, so an exact pin was only holding developers back from newer tooling for no reason. `samtools`/`bcftools` each carry their own tight `htslib >=1.24,<1.25.0a0` run-constraint, so the solver keeps the three coherent regardless.

Conda-vs-PyPI name mappings are called out inline where they differ: `duckdb`→`python-duckdb`, `docker`→`docker-py`, `matplotlib`→`matplotlib-base`, `importlib-metadata`→`importlib_metadata`.

### Verification

`mamba env create --dry-run` solves cleanly — **326 packages** — and now resolves **django 5.2.15** (the patched release) instead of 5.2.5. Every other resolved version lands inside its pyproject bounds: numpy 2.5.1 (<3), pandas 2.3.3 (<3), pydantic 2.13.4 (<3), python-duckdb 1.5.4 (<2), sqlglot 28.10.1 (<29). Python resolves to 3.12.13, and `htslib`/`samtools`/`bcftools` to 1.24 rather than freezing at 1.22 — the floors doing their job.

## 2. Remove obsolete `requirements.txt`

Dead file. Nothing references it — no Jenkinsfile, Dockerfile, README, script, or conda recipe, in this repo or anywhere else in the stack (grepped both). It was superseded by `environment.yml` + `dev-environment.yml` when 2828d031a (GPF-817) split the conda environment into base and dev, and was simply left behind.

It had drifted far past stale:

| | `requirements.txt` (dead) | reality |
|---|---|---|
| python | `3.11` | 3.12 |
| django | `4.2.17` | `>=5.2,<5.3` |
| lark | `lark-parser=0.6.7` | `lark 1.2.2` |
| pyarrow | `16.1` | `>=18` |

Its stale `django=4.2.17` (an EOL series) reads as a security finding to anyone grepping the repo, even though Dependabot correctly ignores the file because nothing consumes it. Deleting it removes the false signal.

## Not in scope

`dev-environment.yml` has the same exact-pin design and is likely drifting too, but it maps to the `[dependency-groups]` rather than `[project.dependencies]` and is worth its own pass.


environment.yml
```diff
@@ -2,68 +2,96 @@
...
+  - importlib_metadata>=8.6
+  # web_api [project.optional-dependencies] -> mysql
+  - mysqlclient>=2.2
+
+  # --- conda-only: no pyproject counterpart ---
+  # htslib/samtools/bcftools are the system binaries behind pysam;
+  # pip installs vendor them into the wheel, conda does not.
   - htslib=1.22
   - samtools=1.22
   - bcftools=1.22
```
lchorbadjiev: this should become:
- htslib >= 1.22
- samtools >= 1.22
- bcftools >= 1.22

lchorbadjiev: Done in 76d331e0c — now `htslib>=1.22` / `samtools>=1.22` / `bcftools>=1.22`.

Worth recording why it's safe: samtools and bcftools each declare their own tight htslib run-constraint (`htslib >=1.24,<1.25.0a0` today), so the solver keeps the three on one coherent series regardless — our floor only sets a lower bound. The dry-run now resolves all three at **1.24** rather than freezing at 1.22.

environment.yml
```diff
@@ -2,68 +2,96 @@
...
+# tooling, build pins) and are flagged as such.
 
 name: gpf
 channels:
   - conda-forge
   - bioconda
   - iossifovlab
 dependencies:
+  # requires-python
   - python=3.12
```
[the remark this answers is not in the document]
lchorbadjiev: Good catch — the header did overclaim. Resolved by taking your second option: `python=3.12` stays and the header now names it as the one deliberate exception, with the reason.

Relaxing it to `>=3.12` would be wrong here. Every pyproject says `requires-python = ">=3.12"`, but the project targets 3.12 specifically — CI builds and tests on it, and `web_api/Dockerfile.production`'s wheels-only invariant assumes cp312 wheels. A `>=3.12` floor would let the solver take the newest interpreter and silently diverge this env from what CI actually tests.

One nuance that makes the exception cheap: conda's `=` is a prefix match, so `python=3.12` already accepts every 3.12.x patch — the dry-run resolves 3.12.13. It was never exposed to the rot that bit `django=5.2.5`, which froze an exact patch.

